PR #4, opened 2021-06-15
added what I think that might work for Webster

colourful-land:


nicktaras:
Thanks for this information @colourful-land - It's been useful to help develop the solution. 

I'll now close this.

PR #55, opened 2021-08-29
refactored to TS, updated error-handlers, added logger()

oleggrib:


nicktaras:
@oleggrib the code looks good. Would you be able to run the hotel bogota demo using these changes (switch out the current negotiator for the typescript dist version).

To make this process easier, as Weiwu suggested, let's split the examples into a separate repo inside Github. Which will be closer the actual development process our consumer developers use.

nicktaras:
@oleggrib - I've test this through adding a webpack configuration to allow it to be installed into web applications.

Will now merge.

PR #140, opened 2022-03-15
Feature/auth handling

micwallace:
Hey @nicktaras,

These commits aim to move integration logic out of the authenticator.js library and into TN.

In detail:
1. Move attestation postMessage functionality out of authenticator.js library and into auth-handler.ts. 
2. Create a mechanism to show iframe modal in client when requested by attestation.id. Outlet will make the iframe visible and pass a response back to the client to show its iframe in a modal.

I think a lot of this functionality could be consolidated in the messaging.ts class eventually but this is a good intermediate step, as it allows errors from attestation.id to flow back to the client and show in the popup.

### DIFF
--- a/src/client/index.ts
+++ b/src/client/index.ts
@@ -415,16 +415,16 @@ export class Client {
                 "<small>You may need to sign a new challenge in your wallet</small>"
             );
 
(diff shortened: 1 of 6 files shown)


micwallace:
Seems origin is actually checked in postMessageAttestationListener.

PR #141, opened 2022-03-16
Fix/token scrolling

micwallace:
Hey @nicktaras,

This fixes the issue with scrolling position on the token views.


micwallace:
PS: I have also gitignored the dist folder.

PR #148, opened 2022-03-29
Refactor/attestation js package

micwallace:
Hey @nicktaras,

This pull request implements the attestation package for use in the outlet & negotiator. 
Currently only the getUseTicket function is used in the outlet but it would be a good place to put some validation 
functions for the client to validate the returned proof (I have started playing around with this on a separate branch). 

Cheers,
Michael


micwallace:
@nicktaras This is ready for merge.

Issue #65, opened 2021-10-06
Set up Automatic Semantic Versioning + Publish Step inside CI Config


nicktaras:
This task should include:

- npx semantic-release-cli setup
- following steps to configure the package
- add documentation for other developers to install the package, access tokens, etc.
- add a publish step to the actions CI, this should build > test > publish to NPM

See: https://www.youtube.com/watch?v=QZdY4XYbqLI

samwarwick:
For further reference: https://docs.npmjs.com/about-semantic-versioning

nicktaras:
Closing this ticket. 

- plans for automatic semantic release will be paused for now
- NPM package discloses information how to get started/also see examples https://github.com/TokenScript/token-negotiator-examples

PR #160, opened 2022-04-14
Added ticket proof validation to client.

micwallace:
Hey @nicktaras 

This PR adds ZK proof validation into the TN client authenticate function.

### DIFF
--- a/package.json
+++ b/package.json
@@ -26,7 +26,7 @@
   },
   "homepage": "https://github.com/TokenScript/token-negotiator#readme",
   "dependencies": {
-    "@tokenscript/attestation": "^0.3.1",
+    "@tokenscript/attestation": "^0.3.3",
     "@toruslabs/torus-embed": "^1.20.2",
     "@walletconnect/web3-provider": "^1.7.1",
     "asn1js": "^2.0.26",
--- a/package.json
+++ b/package.json
@@ -26,7 +26,7 @@
   },
   "homepage": "https://github.com/TokenScript/token-negotiator#readme",
   "dependencies": {
-    "@tokenscript/attestation": "^1.0.7-alpha",
+    "@tokenscript/attestation": "^0.3.1",
     "@toruslabs/torus-embed": "^1.20.2",
     "@walletconnect/web3-provider": "^1.7.1",
     "asn1js": "^2.0.26",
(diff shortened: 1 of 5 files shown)


package.json
```diff
@@ -26,15 +26,15 @@
   },
   "homepage": "https://github.com/TokenScript/token-negotiator#readme",
   "dependencies": {
-    "@tokenscript/attestation": "^1.0.7-alpha",
+    "@tokenscript/attestation": "^0.3.1",
```
samwarwick, line 29: The versioning scheme has been puzzling me. I noticed earlier today at https://www.npmjs.com/package/@tokenscript/attestation that the numbering had seemingly gone backwards. Is this intentional/by-design?

nicktaras: Hi @samwarwick, GP. Let's keep to one approach.

It seems in this case the method of versioning has changed (from alpha .... which would then start at 1.0.0 once production ready / to 0.3.1 which will work towards 1.0.0). There are semver tools that can prompt us of the type of changes are being added e.g. minor, major / breaking changes, bug fix. There's an issue inside Git to implement this before we get to being prod ready.

samwarwick: Adding the existing Token Negotiator ticket for reference: https://github.com/TokenScript/token-negotiator/issues/65. Might need something similar for Attestation.

micwallace: > The versioning scheme has been puzzling me. I noticed earlier today at https://www.npmjs.com/package/@tokenscript/attestation that the numbering had seemingly gone backwards. Is this intentional/by-design?

Hi @samwarwick this was done to keep the major/minor version in sync with the Java attestation library.

micwallace:
@jot2re You may want to have a look at this too: 

https://github.com/TokenScript/token-negotiator/pull/160/files#diff-24f1ac19d22913269d99a8a5afa4c13aa7fd2c8b9858529046424841ca6421daR403

Uses code from attestation branch here:
https://github.com/TokenScript/attestation/blob/e828eca3dcf36bf57fcbf95a93310b94b06f865a/src/main/javascript/crypto/src/Authenticator.ts#L130

jot2re:
Once the feedback on[ Attestation PR 239](https://github.com/TokenScript/attestation/pull/239 ) gets handled, then it looks fine to me

nicktaras:
> 

@micwallace, please feel free to fix any src/client/index.ts conflicts and merge this when ready (Attestation PR is ready etc).

PR #162, opened 2022-04-17
Feature/css fixes

nicktaras:


nicktaras:
CSS fixes added from testing with 3rd party libs (Alchemy React boiler plate Dapp) which obscured the css. These changes enforce the correct styling with a higher level of specificity and additional rules.

PR #165, opened 2022-04-19
Feature/api config

micwallace:
Hey @nicktaras 

This branch adds some changes to on-chain token module to make it easier to configure & to enhance error handling. 

- Move API paths & keys config into a config array. 
- Abstract API call logic to remove duplicate code blocks. 
- Throw error on non successful HTTP error codes. 
- Ensure functions always returns a value when errors are encountered.


nicktaras:
@micwallace - let's not merge this PR as it contained some duplication work / features from another PR. 

I've taken the code from this and the other and merged to:

https://github.com/TokenScript/token-negotiator/pull/169

micwallace:
@nicktaras I agree but please try to avoid manually copying changes from other branches. I merged this branch into #169 and there were a few things that were missed and broken during the merge. 

And if the merge is done through git it maintains commit history for files & branches rather than the changes seemingly coming from nowhere when done manually. 

Also I will continue to spruke Jetbrains IDE since it's git integration makes complex merges like this so much easier :-)

Issue #172, opened 2022-04-22
Outlet/Client ticket decoder


oleggrib:
Currently user have to pass SignedDevconTicket decoder to the Outlet/Client.
Its not included to the repo, but its required.

My suggestion is to embed default SignedDevconTicket decoder and let user option to pass own decoder.

Current SignedDevconTicket is done in bit unusual way (in comparison to Authenticator code), so we can use code from Authenticator or better even import code from Authenticator. (we can do it in future optimization)

PR #173, opened 2022-04-23
added listener to renegotiate tokens when new token added, updated ti…

oleggrib:
added listener to renegotiate tokens when new token added, updated ticket parser to universal, allowed Outlet to work in parent window, added few comments to manual, disabled request to connect to Metamask on negotiate()

@nicktaras , @micwallace , can you check if it looks ok for you?

### DIFF
--- a/README.md
+++ b/README.md
@@ -38,6 +38,20 @@ Within your application install the token negotiator:
   npm i @tokenscript/token-negotiator
 ```
 
+### In case if you want to use negotiator directly then you have to build UMD script 
+
+```npm run build-umd```
+
+and include _negotiator.js_ from _dist_ folder to the HTML page
+
+```<script type="text/javascript" src="./negotiator.js"></script>```
+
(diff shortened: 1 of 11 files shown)


oleggrib:
resolves https://github.com/TokenScript/token-negotiator/issues/172

nicktaras:
Thanks @oleggrib, the changes in the TN look good. I'll test this on Monday and merge if all is working. 

I will just need to ensure the other examples work where the outlet is seperate from the token outlet e.g. hotel bogota.

I'm not entirely across the changes to the DevconTicket files - will we need to re-generate mock tokens once these changes have been merged?

micwallace:
@oleggrib looks fine to me. Although I would like to see all asn schema stuff moved into the attestation library in the long run. The schema code is already there, we just need to add a library interface for decoding schemas, either from a JSON object, or as a fallback, using inbuilt SignedDevconTicket.

micwallace (approved):

PR #178, opened 2022-04-23
Merge test

nicktaras:


nicktaras:
Screen Shot 2022-04-24 at 1 24 27 pm

@oleggrib, I'm looking into this error inside the Passive Devcon Website example.

README.md
```diff
@@ -38,6 +38,20 @@ Within your application install the token negotiator:
   npm i @tokenscript/token-negotiator
 ```
 
+### In case if you want to use negotiator directly then you have to build UMD script 
+
```
nicktaras, line 42: I've added a section from line 243 to explain the use of TN UMD module. Based on feedback from a user of TN, they did not know what UMD is and may confuse developers.

PR #180, opened 2022-04-27
Fix/nextjs support

micwallace:
Hey @nicktaras 

This branch adds these things: 

1. Removed CSS import within token negotiator. Many frameworks, including next.js don't allow CSS imports from within node modules. Examples have been updated to import CSS on wd-example branch. 
2. On chain authentication challenge signing and and address signature check. 
3. Fab overlay displays loader & message when authenticating in order to improve UX.


micwallace:
@jot2re Can you have a quick look at this. It's getting UN from the same place as off-chain authentication (crypto verify server) and getting the user to sign a message with it. It then verifies the signature on the frontend. The signature can then be sent to Webster's backend for verification, along with NFT ownership check: 
https://github.com/TokenScript/token-negotiator/blob/14c024507480f26adbe077c61965bdf668d25173/src/client/index.ts#L336

Websters server should get the current UN from the crypto verify service independently, otherwise the frontend can specify any signed challenge.

jot2re:
The code seems reasonable to me for the flow where a third party webserver wants to verify that the user owns a specific token (that exists on the blockchain).

package.json
```diff
@@ -1,6 +1,6 @@
 {
   "name": "@tokenscript/token-negotiator",
-  "version": "1.0.18-alpha",
+  "version": "1.0.19-alpha-wd",
```
nicktaras, line 3: Can we remove the -wd from the version. Let's move to using Semantic versioning.

https://docs.npmjs.com/about-semantic-versioning (applying these rules to future updates)

src/client/index.ts
```diff
@@ -14,7 +14,6 @@ import {
 } from "../tokenLookup";
 import OnChainTokenModule from "./../onChainTokenModule";
 import Web3WalletProvider from "./../utils/Web3WalletProvider";
-import "./../theme/style.css";
```
nicktaras, line 17: Thanks for providing more detail around the issues from the Next JS forum about this (https://github.com/vercel/next.js/issues/19936) - the reasoning makes sense, even though it incurs more work for web developers.

To implement this change can you update the ReadMe with steps for the developers to install the CSS alongside the core lib.

src/client/index.ts
```diff
@@ -25,6 +24,7 @@ interface NegotiationInterface {
     filters: {};
   };
   onChainKeys?: { [apiName: string]: string };
+  ipfsBaseUrl?: string
```
nicktaras, line 27: Good idea :) Especially if we find people want to offload requests onto their servers.

package.json
micwallace: Yep this was only so Feng can build the brand connector demo.

src/client/index.ts
micwallace: @nicktaras yes, readme now updated.

src/client/index.ts
micwallace: Yeah it make things a lot faster when using Moralis/Alchemy, which returns raw metadata that often includes ipfs links using public gateways. This is different from Opensea which returns a cached CDN URL.

PR #181, opened 2022-04-27
Fix tab messaging for safari browser.

micwallace:
@nicktaras small patch to fix Safari support. 

Since this is always a restriction in safari I decided to check for Safari directly rather than use the local storage check. 
Other browsers still use the iframe cookie check in case they're using stricter privacy controls.


nicktaras:
Thanks @micwallace - just tested this through and looks good mate.

PR #210, opened 2022-05-17
Use schemas from attestation library

micwallace:
- Remove ASN schemas from TN.
- Update references to use schemas from attestation lib.
- Remove asn1-schema dependency from TN.


src/core/__tests__/core.spec.ts
```diff
@@ -93,7 +93,7 @@ describe('core Spec', () => {
 			rawTokens,
 			readSignedTicket,
 			unsignedTokenDataName
-		)).toEqual([{"devconId": "9", "ticketClass": 12, "ticketIdNumber": 1}]);
+		)).toEqual([{"devconId": "9", "ticketClass": 12, "tiketIdNumber": 1}]);
```
nicktaras, line 96: @micwallace - there seems to be a typo added here for ticket. Is this intended?

PR #216, opened 2022-05-19
added fix to format api settings for issuers

nicktaras:


nicktaras:
- Added logic to format issuer naming to lowercase with prefixed hyphens when spaces are added.
- Lint code
- Added unit tests to adjusted functions
- console warning added to inform the web developer when the collectionID is formatted (uppercase of spaces used)

micwallace (approved):

src/client/index.ts
```diff
@@ -104,35 +95,57 @@ export class Client {
 		this.messaging = new Messaging();
 	}
 
+	formatCollectionID(collectionID: string) {
+		let formatedCollectionID = collectionID;
+
+		if (/[A-Z]+/g.test(collectionID) || /\s+/g.test(collectionID)) {
+			formatedCollectionID = collectionID.replace(/\s+/g, "-").toLowerCase();
```
oleggrib, line 102: if its important to us to stay only low-case letters, digits and "-" then we have to replace/remove all special symbols too.
is this allowed for example? "df&`2?\n\t*}}"
so better stay only allowed symbols, not just lowercase them.
@micwallace , @nicktaras

PR #242, opened 2022-06-03
Feature/safe connect

micwallace:
Opening this PR to prep for merge once safe connect is launched.


micwallace:
@nicktaras This is ready for review & merge.

Here the overview:

- Add safe connect as a mock wallet connection. It is optionally enabled via safeConnectOptions client params and supports current authentication methods (zkproof & UN signing). 
- Optionally, a safe connect proof type can be requested during wallet connection and stored for later calls to authenticate. This is done so that there is no need to redirect the user back to safe connect to sign the challenge once "authenticate" is function is called later (unless of course the proof has expired). 
- Implement modular architecture for proof/authentication methods. Current authentication methods remain the default. The developer can specify which proof they require when calling authenticate by using the new type & options fields on  AuthenticateInterface.

PR #283, opened 2022-07-02
Feature token discovery api

nicktaras:


nicktaras:
Changes:

- Updated UMD Output from **token-negotiator-alpha-dist** to **token-negotiator-dist** & ReadMe
- Removed **onChainModule** and related unit tests
- Added **token/nftProvider** to Token Discovery API
- Unit tests added for two methods to connect with Token Discovery API
- Adjusted { client } module to consume the **nftProvider** instead of onChainModule

nicktaras:
- Added silent http error option, to allow the TN to load when collection items are not found (e.g. 500 from API).

micwallace:
@nicktaras I was wondering if we should be removing the other token sources completely. This change means the system relies solely on the uptime of the discovery API and there is no fallback. Maybe we should keep at least one fallback?

nicktaras:
> @nicktaras I was wondering if we should be removing the other token sources completely. This change means the system relies solely on the uptime of the discovery API and there is no fallback. Maybe we should keep at least one fallback?

Thanks good point to raise, I've reached out to Hu to better understand the stability of the API. Let's wait for his response and decide from there.

nicktaras:
Update to status, reached out to Hu and Miccy to confirm for this current PR we can progress with the full migration to Token Discovery - moving this into staging for test / also include some stability testing before we migrate.

micwallace (approved):

PR #293, opened 2022-07-11
Inline UI and async contract metadata load

micwallace:
This PR adds the following features: 

- New inline UIType option inserts TN UI in an element of the developers choosing. 
- Contract metadata is now loaded asynchronously to prevent it from blocking UI initialization. 
- UI functions are exposed through client getUI function. This allows the developer to utilize TN's UI for custom loading and error messages.


nicktaras (approved):
Hi @micwallace, looks good to merge into staging.

PR #294, opened 2022-07-11
Fix auth return values

micwallace:
Hey @nicktaras,

Proposed changes to authenticate return type. What do you think? AuthenticationResult interface also has type and target properties, but these are for internal use and mostly irrelevant to developers.


src/client/index.ts
```diff
@@ -413,8 +413,8 @@ export class Client {
 		emitSelectedTokensToClient: (tokens: any) => {
 			this.on("tokens-selected", null, { selectedTokens: tokens });
 		},
-		emitProofToClient: (proof: any, issuer: any, error = "") => {
-			this.on("token-proof", null, { proof, issuer, error });
+		emitProofToClient: (data: any, issuer: any, error = "") => {
+			this.on("token-proof", null, { data, issuer, error });
```
nicktaras, line 417: Looks good

PR #298, opened 2022-07-12
Proposed NegotiateInterface changes.

micwallace:
Hey @nicktaras, since next version will most likely be a major release, I thought it'd be the best time to propose some changes to the config interface. I know you've been meaning to move filters to the issuer config, so I've done that here and moved overlay config to uiOptions. Let me know what you think. If you will agree I'll apply these updates to the readme tomorrow.


nicktaras:
@micwallace 

> Hey @nicktaras, since next version will most likely be a major release, I thought it'd be the best time to propose some changes to the config interface. I know you've been meaning to move filters to the issuer config, so I've done that here and moved overlay config to uiOptions. Let me know what you think. If you will agree I'll apply these updates to the readme tomorrow.

Yep, these breaking changes make sense to add now, in relation to adding the embedded UI feature (as part of the e-commerce solution coming into place soon).

There's some work for us to do to ensure these changes align with current projects and examples + Readme. Thanks, if you can update the Readme to align with these changes as part of the PR.

micwallace:
@nicktaras Readme updated with configuration reference table. 

I also moved autoPopup to uiOptions.

micwallace:
Hey @nicktaras 

I've made a few more changes and updated the readme with a configuration reference. The reference has all options available for client & outlet.

I've left out some Outlet related settings from the readme as they are likely redundant or rarely needed at this time. Now rather than these options being required, they have defaults set in the outlet class. Also, all outlet-only options have been removed from the client's OffChainTokenConfig interface.

nicktaras:
> Hey @nicktaras
> 
> I've made a few more changes and updated the readme with a configuration reference. The reference has all options available for client & outlet.
> 
> I've left out some Outlet related settings from the readme as they are likely redundant or rarely needed at this time. Now rather than these options being required, they have defaults set in the outlet class. Also, all outlet-only options have been removed from the client's OffChainTokenConfig interface.

Thanks Miccy. I agree with your changes to flatten the config structure (breaking change etc). Let's see if Oleh has any suggestions / feedback for this change.

src/outlet/index.ts
```diff
@@ -32,7 +51,7 @@ export class Outlet {
 	urlParams?: URLSearchParams;
 
 	constructor(config: OutletInterface) {
-		this.tokenConfig = config;
+		this.tokenConfig = Object.assign(defaultConfig, config);
```
oleggrib, line 54: @nicktaras , this assign do single level assign, its not recursive, so be careful here, better assign last-level propertires or assign from root to the end.

src/client/interface.ts
```diff
@@ -33,21 +23,21 @@ export interface IssuerConfigInterface {
...
-    };
-    ipfsBaseUrl?: string;
+    uiOptions?: UIOptionsInterface;
     autoLoadTokens?: number | boolean;
     autoEnableTokens?: boolean;
-    autoPopup?: boolean;
     messagingForceTab?: boolean;
     safeConnectOptions?: SafeConnectOptions;
+    unSupported?: {
+        config: BrowserDataInterface,
```
oleggrib, line 38: @nicktaras , "config" property name is not obvious. can we rename it to "browserList" ?

oleggrib (commented):
I have added comment and commit to update README, other changes looks good

src/client/interface.ts
nicktaras: Thanks @oleggrib, to make this clear to the web developer, this will be updates to "unSupportedUserAgent"

src/outlet/index.ts
nicktaras: Thanks @oleggrib - checked in with @micwallace on this, his changes applied are a single level deep from reviewing this together. We will proceed with this to optimise browser performance (ensuring we build update this within flat interface / no nested properties etc).

nicktaras (approved):

src/client/index.ts
```diff
@@ -24,14 +24,11 @@ declare global {
...
 	issuers: [],
-	options: {
-		overlay: {
-			uiType: "popup",
-			containerElement: ".overlay-tn",
-			openingHeading: "Validate your token ownership for access",
-			issuerHeading: "Detected tokens"
-		},
-		filters: {}
+	uiOptions: {
```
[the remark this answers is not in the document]
micwallace: All the options there are related to the UI. So it makes sense to group them together. Any other options should be top level unless they are related.

PR #325, opened 2022-07-20
FEAT: Added check for non supported browsers/userAgents

BankiLouis:
This PR has updates for checking nonsupported browsers/platforms.


micwallace (changes requested):
Hey @BankiLouis 

Good work on your first PR for negotiator. 
There is one thing in these changes that I think should be optimized:

The functions checkBrowserSupport & isUserAgentSupported are very similar. Rather than having checkBrowserSupport in Client.ts, merge it with the existing isUserAgentSupported function in isSupported.ts. isUserAgentSupported is already used in Client for the authenticate function.

nicktaras:
> Hey @BankiLouis
> 
> Good work on your first PR for negotiator. There is one thing in these changes that I think should be optimized:
> 
> The functions checkBrowserSupport & isUserAgentSupported are very similar. Rather than having checkBrowserSupport in Client.ts, merge it with the existing isUserAgentSupported function in isSupported.ts. isUserAgentSupported is already used in Client for the authenticate function.

Really point @micwallace.

We haven't release 2.0 yet, giving an opportunity for us to re-work the unSupported config.

How about we, adjust the following structure of the new TN config from:

````
client = {
unSupportedUserAgent: {
        config: {
            // options to test against : ["iE", "iE9", "edge", "chrome", "phantomJS", "fireFox", "safari", "android", "iOS", "mac", "windows", "webView", "touchDevice", "metaMask", "alphaWallet", "mew", "trust", "goWallet", "status", "imToken", "metaMaskAndroid", "alphaWalletAndroid", "mewAndroid", "imTokenAndroid"];
            metaMaskAndroid: true,
            alphaWalletAndroid: true,
            mewAndroid: true,
            imTokenAndroid: true,
        }
    }
}
````

to:

````
client = {
unSupportedUserAgent: {
        authentication: {
            config: {
                metaMaskAndroid: true,
                alphaWalletAndroid: true,
                mewAndroid: true,
                imTokenAndroid: true,
            },
           warningMessage: '....'
        },
        full: {
            config: {
                iE: true,
                iE9: true,
            },
           warningMessage: '....'
        }
    }
}
````

This should allow us to use the same logic for both use cases.

micwallace:
Is there any reason to have it as an option? Why not hard-code it if we are not supporting IE in any circumstances.

BankiLouis:
> Hey @BankiLouis
> 
> Good work on your first PR for negotiator. There is one thing in these changes that I think should be optimized:
> 
> The functions checkBrowserSupport & isUserAgentSupported are very similar. Rather than having checkBrowserSupport in Client.ts, merge it with the existing isUserAgentSupported function in isSupported.ts. isUserAgentSupported is already used in Client for the authenticate function.

Thanks @micwallace, yeah I think that will be ideal. Let me get on that.

micwallace (changes requested):

micwallace (approved):

PR #327, opened 2022-07-24
updated support interface and unit tests to include a clear definitio…

nicktaras:
…n for the end user of this feature. useragent > type > config and warning.


nicktaras:
Please review this slight change to fully define the use case in the API for the user agent.

Before: unsupportedUserAgent > config & message
Now: unsupportedUserAgent > type > config & message

This giving us the ability to extend this with other unsupported userAgent related features any time in the future.

micwallace (approved):

PR #329, opened 2022-07-24
minimise contents of NPM package

nicktaras:


micwallace (changes requested):
package.json definitely needs to be included in the NPM package

micwallace:
PS: This branch should be refactor/ or chore/ as it's not a feature branch.

nicktaras:
> package.json definitely needs to be included in the NPM package

Removed package.json from npmignore.

micwallace (approved):

PR #336, opened 2022-07-28
FEAT: Added check and error handling for internet connectivity.

BankiLouis:
This PR has updates for internet connectivity error. Added `navigator.offline` listener in the negotiate function to handle internet disconnection error.


src/client/index.ts
```diff
@@ -538,10 +540,24 @@ export class Client {
...
+	checkInternetConnectivity(): void {
+		if (!navigator.onLine) {
+			if (this.config.type === 'active') {
+				setTimeout(() => {
+					this.ui.showError(this.config.noInternetErrorMessage ?? NO_INTERNET_ERROR_MESSAGE);
+				}, 1000);
+			}
+			throw new Error(this.config.noInternetErrorMessage ?? NO_INTERNET_ERROR_MESSAGE)
+		}
+	}
```
nicktaras, line 559: I think we should add a return boolean value to this function. This so we can use this inside the logic to stop the following action from triggering if there is no connection.

src/client/views/start.ts
```diff
@@ -20,7 +20,8 @@ export class Start extends AbstractView {
 		this.viewContainer.querySelector('.opening-btn-tn').addEventListener('click', this.goToWalletSelection.bind(this));
 	}
 
-	goToWalletSelection(){
+	goToWalletSelection() {
+		this.client.checkInternetConnectivity();
```
nicktaras, line 24: I don't think the following functions should trigger if there is no internet found at that time. This should help smooth the UX. Where from testing, this allows the user to continue, but the requests to collect tokens have failed leaving a black screen.

nicktaras:
@BankiLouis the implementation is great for users to manage the message. I've added a couple of comments to review. Thanks.

nicktaras (approved):
Looks good @BankiLouis. Lets merge this in to staging, once we have the next release out. 

Thanks

PR #337, opened 2022-07-29
Fix config merge

micwallace:
I unfortunately finished this commit before I realized that @oleggrib had fixed the same thing.

Let's benchmark both of these and let the most efficient be merged :-)

### DIFF
--- a/src/client/index.ts
+++ b/src/client/index.ts
@@ -93,9 +93,7 @@ export class Client {
 
 	constructor(config: NegotiationInterface) {
 
-		config.uiOptions = {...defaultConfig.uiOptions, ...config?.uiOptions}
-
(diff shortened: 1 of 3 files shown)


nicktaras:
@micwallace / @oleggrib - I was unsure from the last comment if the solution already inside the staging branch, or you've made two seperate solutions and want to merge one in. However, If you guys want to review and merge this together  between the both solutions that would be great.

nicktaras:
@oleggrib and @micwallace - the changes made are like for like with regards to the loop used to migrate the config from the defaults and the users settings. 

I'm going to merge this branch, then look for any further changes needed from Oleh's that improve the logic are differ / providing changes this branch didn't cover.

PR #344, opened 2022-08-05
Feature/accessibility screen readers

BankiLouis:


micwallace (approved):

src/client/views/start.ts
```diff
@@ -10,7 +10,7 @@ export class Start extends AbstractView {
               <div class="brand-tn"></div>
               <div class="inner-content-tn">
                 <div class="inner-content-block-tn">
-                  <button class="opening-btn-tn">Let's go!</button>
+                  <button class="opening-btn-tn" aria-label="Start negotiator button">Let's go!</button>
```
nicktaras, line 13: How about we use 'click to begin connecting your tokens'

nicktaras: Token Negotiator or Brand Connector is not ideal because we use these names interchangeably.

src/client/ui.ts
```diff
@@ -20,7 +20,7 @@ export interface UIOptionsInterface {
 export class Ui {
 
 	private static UI_CONTAINER_HTML = `
-		<div class="overlay-content-tn">
+		<div class="overlay-content-tn" aria-label="Token negotiator popup">
```
nicktaras, line 23: Could we change this to 'token negotiator overlay'.

nicktaras (approved):

PR #366, opened 2022-08-30
added feature to get a connected wallet instance by index

nicktaras:


src/wallet/Web3WalletProvider.ts
```diff
@@ -27,8 +27,21 @@ export class Web3WalletProvider {
 	constructor(client: Client, safeConnectOptions?: SafeConnectOptions) {
 		this.client = client;
 		this.safeConnectOptions = safeConnectOptions;
+		this.getConnectionByIndex = this.getConnectionByIndex;
 	}
 
+	getConnectionByIndex(connectionIndex:number){
+		let index = 0;
```
micwallace, line 34: Here's how I would solve this:

```
if (connectionIndex >= this.connections.length) return null;

return this.connections[Object.keys(this.connections)[connectionIndex]];
```

PR #369, opened 2022-09-01
added wallet provider to tkn dev interface

nicktaras:


nicktaras:
Added feature to share the wallet provider to the end developer using our library.

The use case for this is to enable users and ourselves the ability to trigger web3 events from outside of Token Negotiator such as;

- Signing
- Smart Contract events
- Triggering any of the provider methods custom to an applications use case

micwallace (approved):

PR #370, opened 2022-09-05
added logic to hide solana wallet when object is not present or no is…

nicktaras:
Added logic to hide Solana Wallet Selection when wallet is not required:

- when window.solana is not defined
- when there are no issuers that require solana blockchain


micwallace (approved):
@nicktaras looks good mate!

PR #375, opened 2022-09-07
Feature/solana aligment with api

nicktaras:


nicktaras:
@micwallace could you help review this, where you've had the most time looking at this item

I've re-tested Solana and EVM tokens to ensure functionality is kept.

Where for this task, we will need to merge the Solana Token Discovery API changes to enable this change to best work with Solana.

PR #377, opened 2022-09-11
Update/enhance typescript support and integration

BankiLouis:
This PR fixes the error `Cannot find name 'Web3WalletProvider'.` when integrating with Angular and Also updates on interfaces for better Typescript integration.


src/client/index.ts
```diff
@@ -251,7 +258,7 @@ export class Client {
 
 		if (this.config.type === "active") {
 
-			this.issuersLoaded = !this.tokenStore.hasUnloadedIssuers();
+			this.issuersLoaded = false;
```
nicktaras, line 261: I'll review this in a bit more detail - to ensure that at this point the setting should be false.

PR #382, opened 2022-09-15
feature/nav-css-fix

nicktaras:


nicktaras:
Screen Shot 2022-09-15 at 12 52 29 pm

Applied CSS to stop the root level element nav css, margin or padding being applied to the tn <nav> element.

micwallace (approved):

PR #385, opened 2022-09-15
chore: 🤖 remove dist/ from repo

abcfy2:
Remove `dist/` from repo, this will generate by `npm run build` job.

And I replace `prepublishOnly` task with `prepare`, which will automatically run after `npm ci/install`, and run before `npm pack/publish`. See: https://docs.npmjs.com/cli/v8/using-npm/scripts#prepare-and-prepublish

So both `npm publish` and `npm install "git://github.com/tokenScript/token-negotiator.git#staging"` should work.

And I set `"files": ["dist/"]` in `package.json`, which will limit npm package only contains `dist/` directory and `package.json`.

After all, this should fix CI https://github.com/TokenScript/token-negotiator-examples/pull/96

### DIFF
--- a/package.json
+++ b/package.json
@@ -4,13 +4,14 @@
   "description": "Token-negotiator a token attestation bridge between web 2.0 and 3.0.",
   "main": "dist/index.js",
   "types": "dist/index.d.ts",
+  "files": ["dist/"],
   "scripts": {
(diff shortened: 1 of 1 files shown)


nicktaras:
@abcfy2 thanks. Let's test this out 👍

PR #403, opened 2022-09-26
fix: 🐛 fixed code to work with firefox and chrome and dont fail

oleggrib:
Fixed view (content was invisible)
content added in correct way (original page was broken)
disabled storage API request for browsers, which doesnt support it
updated ticket name (new tickets use _ticket_ , not _tiket_)
cc @nicktaras

### DIFF
--- a/src/outlet/index.ts
+++ b/src/outlet/index.ts
@@ -153,7 +153,9 @@ export class Outlet {
 					// MagicLinkReader part of Outlet usually works in the parent window, same as Client, so it use same document
 					document.body.dispatchEvent(event);
 
-				} catch (e){}
+				} catch (e){
+					// no-op
+				}
 				
 				this.sendTokens(evtid);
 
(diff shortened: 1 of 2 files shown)


src/outlet/index.ts
```diff
@@ -167,21 +171,28 @@ export class Outlet {
...
+		
 		if (origin === document.location.origin)
-			return;
-
+		return;
+		
 		let accessWhitelist = JSON.parse(localStorage.getItem("tn-whitelist")) ?? {};
-		const storageRequestNeeded = window.parent && (document.hasStorageAccess && !(await document.hasStorageAccess()));
+		// window.parent always defined, so no need to check
+		if (!document.hasStorageAccess) {
```
micwallace, line 182: @oleggrib Even if storage access isn't required, we still need to add the site to the whitelist. This will allow old browsers to bypass the permission dialog.

micwallace (changes requested):

src/outlet/index.ts
oleggrib: old browsers dont have this method _document.hasStorageAccess_ , so whole flow will be ignored. is it ok?

micwallace: No I think we always want to show permission dialog. It's better for security & user privacy. We don't want off-chain tokens to be used as a tracking cookie.

micwallace: @oleggrib In the interest of keeping momentum have committed proposed changes to this branch. If you are okay with this let's get this merged, otherwise let's have a discussion and we can revert if needed.

PR #405, opened 2022-09-27
feature: load tokens directly for same-origin outlets

micwallace:
This PR allows tokens to be loaded directly from local storage when negotiator is used on the same origin as the outlet


micwallace:
@oleggrib Token loading & authentication for local outlet is now implemented in this branch.

Authentication works for both iframe and tab flow.

nicktaras (approved):

PR #410, opened 2022-09-29
fix: button svg icon position fix

snowwhitedev:


index.html
```diff
@@ -11,6 +11,12 @@
       src="./token-negotiator-dist/negotiator.js"
     ></script>
     <link rel="stylesheet" href="./token-negotiator-dist/theme/style.css" />
+    <style>
+      svg {
```
nicktaras, line 15: Hi Henderson, should this be applied to the libraries theme.css.

e.g. Within scope of the container element.

.overlay-tn svg { ... }

snowwhitedev: Hi  @nicktaras Nick  No.
I just added this additional style in index.html to see if my changes in theme.css are working correctly.
You know, svg style is set like above as default in **perion**, it showed tweaked button. :)

nicktaras: I think I understand, you added this logic to mirror your Perion implementation. But then applied all changes needed to theme.css.

I've removed the SVG targetting from the index.html (mock example of TKN), can you confirm that all the styles needed for the svg change are inside theme.css. Then let's merge to staging. Thanks.

PR #413, opened 2022-10-02
Feature/toggle option

nicktaras:
Please see this PR created for the management of the Toggle token on/off functionality.

For each token collections config, "hideToggle" flag can be used.


oleggrib (approved):
works great!
I just added commit to remove toggle block completely if it disable, not only hide it
also I added setting to hide toggle to all devcon6 tokenConfigs (prod, stage)

micwallace (approved):

PR #425, opened 2022-10-20
refactor: fix token detail load issue

snowwhitedev:
┆Issue is synchronized with this [Jira Task](https://smarttokenlabs.atlassian.net/browse/TKN-438) by [Unito](https://www.unito.io)


nicktaras:
@snowwhitedev - good idea on that we don't want users clicking load, when the auto load will make the requests for the end user.

Instead of disabling the buttons, which this cursor effect will only work on desktop, not on mobile - can we instead hide the load buttons or just make them inactive, until the auto load is complete.

snowwhitedev:
> @snowwhitedev - good idea on that we don't want users clicking load, when the auto load will make the requests for the end user.
> 
> Instead of disabling the buttons, which this cursor effect will only work on desktop, not on mobile - can we instead hide the load buttons or just make them inactive, until the auto load is complete.

Yep. Got it. I will try

nicktaras:
> > @snowwhitedev - good idea on that we don't want users clicking load, when the auto load will make the requests for the end user.
> > Instead of disabling the buttons, which this cursor effect will only work on desktop, not on mobile - can we instead hide the load buttons or just make them inactive, until the auto load is complete.
> 
> Yep. Got it. I will try

Thanks @snowwhitedev - please let me know once the changes are in place and we can re-review.

PR #436, opened 2022-10-25
Fix/detail load issue

snowwhitedev:


nicktaras:
Thanks @snowwhitedev - and also for sharing the video. 

From taking a look at the animation it's not as smooth as it could be. Could you show the loading for each collection, instead of the hidden buttons, I think this will more smoothly transition into showing the loading / load buttons that follow.

snowwhitedev:
> Thanks @snowwhitedev - and also for sharing the video.
> 
> From taking a look at the animation it's not as smooth as it could be. Could you show the loading for each collection, instead of the hidden buttons, this will smoothly transition to load buttons that follow.

okay, Got it. Let me try it

nicktaras:
9DF6F201-19B8-43F8-83F7-98C5A383EFD9

great work on this. Can you add a unit test towards the changes added to enable Sonar Cloud to pass.

nicktaras (dismissed):

nicktaras (approved):
Approving. This area needs some work on the unit tests.

PR #437, opened 2022-10-25
Tkn 247 unit test coverage auth

BankiLouis:


nicktaras (approved):

nicktaras:
@BankiLouis - I'll merge this. It looks like the issue is that theres a removal of a dep in a file and there's no unit test etc.

PR #438, opened 2022-10-27
Feature/wallet api hooks

nicktaras:


nicktaras:
Hi guys, can you please review this change which adds a network changed hook the Token Negotiator to notify when a user changes the network used of their EVM wallet. 

A follow up task has been written for a lower priority item, to enable the same functionality for Solana wallet network changes and also changes to the account connected. https://smarttokenlabs.atlassian.net/browse/TKN-466

BankiLouis (approved):

PR #440, opened 2022-10-28
Oh/browser support

oleggrib:
Fixed browser detection
Added Auto-outlet fire for multiple redirect scenatios
Auto-add contrainer if not exists
Added message to iOS brawe owners to close tab

examples for testing in browsers:
https://perks.antopolbus.rv.ua/token/local.php
https://perks.antopolbus.rv.ua/token/remote.php

@jeanhwa could you test in with your devices?
For me it works for all devices.
Issues only with iOS Brave. it cant close child tab and in some conditions localStorage can be disabled.


src/client/index.ts
```diff
@@ -717,9 +730,91 @@ export class Client {
...
+			}
+		})
+
+		return onlySameOriginFlag;
+	}
+
 	on(type: TokenNegotiatorEvents, callback?: any, data?: any) {
 		requiredParams(type, "Event type is not defined");
 
+		if(type == 'token-proof') {
```
nicktaras, line 791: Should this be ===

src/client/index.ts
```diff
@@ -717,9 +730,91 @@ export class Client {
...
+			logger(2, "token-proof listener atteched. check URL HASH for proof callbacks.");
+			let params =
+			window.location.hash.length > 1
+			? "?" + window.location.hash.substring(1)
+			: window.location.search;
+			let urlParams = new URLSearchParams(params);
+			
+			const action = urlParams.get("action");
+						
+			if (action == "proof-callback") {
```
nicktaras, line 801: Should this be ===

nicktaras:
Thanks @oleggrib. 

Can Jean do all initial testing with demo URL’s shared? 

I’ll review the code first thing on Monday morning. If urgent, I can do this earlier. 

I’ve created Jira ticket 470 to help track the progress. Oleh, can you add details to this to the improvements and all change made that Jean can check as he tests.

src/client/index.ts
oleggrib: fixed in https://github.com/TokenScript/token-negotiator/pull/440/commits/534669300597f5cb437aff3cbd7677c37ce6c4e9

src/client/index.ts
oleggrib: fixed in https://github.com/TokenScript/token-negotiator/pull/440/commits/534669300597f5cb437aff3cbd7677c37ce6c4e9

nicktaras:
> Thanks @oleggrib.
> 
> Can Jean do all initial testing with demo URL’s shared?
> 
> I’ll review the code first thing on Monday morning. If urgent, I can do this earlier.
> 
> I’ve created Jira ticket 470 to help track the progress. Oleh, can you add details to this to the improvements and all change made that Jean can check as he tests.

Hi @oleggrib - can I check the action on this ticket. Is Jean to test via the links you’ve shared first? Then we move to staging?

oleggrib:
> Hi @oleggrib - can I check the action on this ticket. Is Jean to test via the links you’ve shared first? Then we move to staging?

yes, you can use my links to test how it works.

nicktaras:
> > Hi @oleggrib - can I check the action on this ticket. Is Jean to test via the links you’ve shared first? Then we move to staging?
> 
> yes, you can use my links to test how it works.

@jeanhwa can you help test these changes, if all looks good we'll merge into staging.

jeanhwa:
@oleggrib can I know what result should I expect from testing? 
I tested with https://perks.antopolbus.rv.ua/token/remote.php and after add ticket to domain and open remote token, console log shows active mode for these browsers below while other browsers show passive mode. is this expected?

windows - brave
mac - safari and brave
android - brave
ios - safari, edge and brave

active

jeanhwa:
In iOS firefox, console log shows popup blocked error

unnamed (1)

cc @oleggrib @nicktaras

nicktaras:
> In iOS firefox, console log shows popup blocked error
> 
> unnamed (1)
> 
> cc @oleggrib @nicktaras

Thanks @jeanhwa 

@oleggrib - is this expected in this change.

oleggrib:
Thank you @jeanhwa . Really Firefox on iOS require active flow
I updated code for browser detection and forced active flow for iOS Firefox

src/client/index.ts
```diff
@@ -276,7 +277,7 @@ export class Client {
 
 			let err = this.config.unSupportedUserAgent[type].errorMessage;
 
-			if (this.config.type === 'active') {
+			if (this.activeNegotiateRequired()) {
```
micwallace, line 280: @oleggrib I don't like this. active/passive should always be a user option. Passive = no-UI (we should probably change the name to suit) as there are some cases where the user wants to implement their own UI. Apart from UI there is really no difference between passive & active.

src/client/index.ts
```diff
@@ -297,6 +312,14 @@ export class Client {
 
 	async negotiate(issuers?: (OnChainTokenConfig | OffChainTokenConfig)[], openPopup = false) {
 
+		let currentIssuer = this.getOutletConfigForCurrentOrigin();
```
micwallace, line 315: @oleggrib Am I correct in thinking that we no longer need to initialize Outlet on pages where Client is used as well? For instance, I can just initialize client and call negotiate and magic link will be read?

micwallace:
@oleggrib @nicktaras Passive flow can work for all browsers, but in passive mode it's up to the user code to detect POPUP_BLOCKED error and show some button that re-triggers negotiation. passive mode is essentially "no UI". Apart from that there's barely any difference.

I suggest we rename this config in future to be more verbose and deprecate the previous values.
type "ui" - old active mode
type "headless" - old passive mode

src/client/index.ts
oleggrib: In case of safari/brave/iOS_FF current flow cant be passive, because it requires USER ACTION to open new tab.
If we want to make it really passive and avoid actions then we have to change flow to request tokens list with REDIRECT_FLOW @micwallace @nicktaras 
what do you think?

src/client/index.ts
oleggrib: right, this is why I did it, because developer can forget to add Outlet code and MagicLInk save can be broken, same as redirect flow. @micwallace

oleggrib:
> @oleggrib @nicktaras Passive flow can work for all browsers, but in passive mode it's up to the user code to detect POPUP_BLOCKED error and show some button that re-triggers negotiation. passive mode is essentially "no UI". Apart from that there's barely any difference.

I think passive move require UI for some browsers and better make our one-style solution to interact with user.
in passive mode we can avoid display tokens, but instead show button to start receive tokens. Its not a best idea to ask webster to code/style some button for some action.
webster (third party developer, who use TN) should not care about TN internal flow and ubderstand error codes and do actions for errors. If webster has to do it then he will be very confused and will not like TN 
@micwallace @nicktaras

oleggrib:
> I suggest we rename this config in future to be more verbose and deprecate the previous values.
type "ui" - old active mode
type "headless" - old passive mode

right now we have few switches in active/passive flow:
1. auto-request tokens or wait for user action to start(wait on active mode)
2. show token list or not (not on passive mode)
3. on browser NEW TAB error show button to open new tab

all of that can be handled in single pop-up, but we can show/hide pop-up depends on content. if content present the show, otherwise hide popup.

so UI can be enabled in both flows, but sometimes hidden

so instead of active/passive flow we can add 2 switches:
1. negotiate_on_action true/false 
2. show_tokens true/false

make sense? @micwallace @nicktaras

src/client/index.ts
nicktaras: Hi guys. 

I think it's a good idea to make the developers life easier. However by allowing this, what's the outcome if the developer does initialise the outlet on a { client, outlet } configuration?

Is it not best to keep this items seperate for now and optimise later?

Thanks guys, open to the direction we take my side.

src/client/index.ts
nicktaras: Hi guys, 

I agree with Miccy on this that we should support the passive flow on all browsers over time.

However if I understand correctly this feature is not currently available for full attestation in any browsers accept for chrome. However all browsers can learn attestations via passive.

What are your thoughts that we remove the full attestation support from passive for the time being? Including the passive example hotel bogota?

Thanks.

micwallace: @oleggrib passive mode can be used in the browsers you mentioned, I have done this for hotel Bogota passive and ticket issuer examples. It just requires that Webster handles the POPUP_BLOCKED exception to show a button that re-triggers negotiate when clicked.

micwallace: @nicktaras this is not related to authentication, just negotiate.

src/client/index.ts
micwallace: @oleggrib @nicktaras I like the idea of having outlet initialised by negotiator for same-origin outlets. 

However there are a few considerations:
1. Do we really need to initialise the whole outlet instance just to read a magic link. 
2. I think magic link reading should be separated from outlet to reduce complexity, maybe a static method that can be called within negotiator, outlet or even directly.
3. From looking at the diff, I can't see the LocalOutlet class being used to read the tokens direct from local storage. Maybe I am just tired LOL

src/client/index.ts
nicktaras: @micwallace - thanks. I can see that the tickets can be learnt in the examples currently with passive. It's the full attestation flow that cannot be completed.

src/client/index.ts
nicktaras: Hi guys, I'm going to close these conversations.

Thanks @micwallace - you're right, we should look to break down the code / optimise further. Where we could have a module named - { magicLink } to include in examples where this is required to lighten the code inside outlet.

For now, let us first focus on the delivery of BC V2, and park up refactoring until later / unless needed during this development phase.

src/client/index.ts
nicktaras: Great inputs on this. 

The outcome is that Oleh is going to create some changes on this PR;

- By default Safari and Brave will auto-redirect for passive flows to acquire token data
- An option will be added to the lib's API for developers to use this default setting, or use a fall back which will use active when passive can't be done which out a ui. This fall back and area can be looked at as a work in progress, where the navigation between tabs will be a key area for improvement for the foreseeable future until we can smooth this flow with less barriers across devices/wallets

This solution will work and enable the flow for devs that use passive flow. Thanks again for the inputs and time put into this.

nicktaras:
@oleggrib - I've resolved the conversations in this PR. Where once the PR is ready please let me know and I can help test this through.

oleggrib:
@nicktaras @micwallace , I updated code. passive flow works without UI POPUP, brave/safari/iOS based on REDIRECT flow for both netotiate() and authenticate()

src/client/index.ts
```diff
@@ -689,6 +750,8 @@ export class Client {
 		}
 	};
 
+	// is it useful? better remove it, we never use it
+	/*
```
nicktaras, line 754: This is required by the ticket issuer site demo.

oleggrib: Better use real user flow for testing to avoid unhandled situations

nicktaras: Resolving this conversation. We've added a work around to the examples ticket issuer that fixes the page flickering. This can be later brought into the core lib if we see this as a common design for use of TKN.

PR #451, opened 2022-10-31
feat: rpc config for wallet connect provider set up

snowwhitedev:


nicktaras:
@snowwhitedev - I've updated the list of RPC's we can support to the Jira Ticket. Could you review and add these. Thanks.

nicktaras (changes requested):
Thanks for the update. 

There are a couple missing from the list including Ethereum.

Please share when ready for next review.

nicktaras (approved):

nicktaras:
Looks good 🔥

PR #471, opened 2022-11-18
feat: modularizing css into scss

snowwhitedev:


nicktaras:
Thanks @snowwhitedev. This looks good, I see that this is a foundation implementation e.g. we haven't switched the actual css around to use mixins, variables etc.

Let one of us test this out on our local machines, then we can bring it into staging.

snowwhitedev:
> Thanks @snowwhitedev. This looks good, I see that this is a foundation implementation e.g. we haven't switched the actual css around to use mixins, variables etc.
> 
> Let one of us test this out on our local machines, then we can bring it into staging.

I  agree. I tested on my local already and attaching short video in the jira

nicktaras:
@snowwhitedev could you take a look at the UMD version, it doesn't seem to work.

Steps:

1. inside the TKN repo run, `npm run build-umd`
2. then run `http-server`

Screen Shot 2022-11-23 at 5 30 56 pm

nicktaras (approved):

nicktaras:
Tested this through and works well for both UMD and NPM package with examples. 

Thanks @snowwhitedev

snowwhitedev:
> Tested this through and works well for both UMD and NPM package with examples.
> 
> Thanks @snowwhitedev

Thank you @nicktaras

PR #475, opened 2022-11-22
feat: reduce time to load of TN on lighthouse performance

BankiLouis:
This PR is to increase the performance metrics of TN on lighthouse analysis. Here are screen shows of before and after optimization
Screen Shot 2022-11-22 at 05 51 54
After
Screen Shot 2022-11-22 at 05 45 45


nicktaras:
Thanks @BankiLouis, this looks good. 

Can you add this example to the ReadMe also to share how this can be installed.

<script defer type="text/javascript" src="./token-negotiator-dist/negotiator.js.gz"></script>

Let one of us run this locally and test it over, then merge into staging.

BankiLouis:
> 

@nicktaras the  negotiator.js.gz script is now part of the webpack build. It is a recommendation by Lighthouse for text compression. So I'm basically just using the compressed file instead. Should I add this detail to the README?

nicktaras:
> 

Yep, there's an example in the Readme. Where you can switch out the code to use the gz example.

snowwhitedev (commented):
When I tried to run it on local, I got this.
screenshot

cc: @nicktaras

BankiLouis:
@nicktaras @snowwhitedev  I've updaed the PR and the changes require the configuration at the server level to be able to serve a gzip compressed version of the negotiator.js. 
using` http-server -g` will server with gzip files which will be picked up by the browser. Below are some screenshots with and without the -g flag on the server.
Screen Shot 2022-11-23 at 10 04 28
Screen Shot 2022-11-23 at 10 04 17
@nicktaras please how can I exclude webpack from sonarCloud web coverage. I have tried the following and it's not working
```
sonar.exclusions=./webpack.config.js
sonar.coverage.exclusions = ./webpack.config.js
```

snowwhitedev:
@BankiLouis 
Yes. it looks pretty good. I am attaching 2 screenshots for desktop and mobile version
desktop
mobile

snowwhitedev (commented):
@BankiLouis @nicktaras I guess if conflict is solved, we can approve and merge this

nicktaras:
Thank you for your time and effort looking at this. The solution that will be used, uses the same solution to you, providing files with Gzip (Gzip on the fly - as coined by GitHub).
Where the files are given in Js format and Githubs NGINX server will do the compression for us. 

I will close this PR, where the solution to do this will be done via our CICD pipeline, solution provided by Hu’s team around this.

PR #476, opened 2022-11-22
added ethers instance inside wallection connection object

nicktaras:


nicktaras:
@snowwhitedev, @BankiLouis - please review and approve when ready.

This change introduces the smart contract interaction library to the wallet connection call back interface. To provide an opinionated way for developers to work with web3 once connected via TKN.

In future we may need to change the libs that we attach here e.g. if they are connected with Solana or Cardano then we can provide the most popular lib for them to work with. 

Thanks guys.

BankiLouis (approved):

PR #487, opened 2022-12-06
Fix: UI init race condition

micwallace:
fix UI race condition when invoking authenticate in token-selected event.


micwallace:
Closing, included in https://github.com/TokenScript/token-negotiator/pull/493

PR #488, opened 2022-12-06
Feature: custom view factory

micwallace:
Using a class (constructor) in the config has the limitation that no external dependencies can be injected into the class, 
since the ViewConstructor<ViewInterface> has a static constructor.

This updates the option to use a factory pattern to allow external dependency injection into overridden views. 

It would be possible, but not future proof to allow the config to accept a factory function and a class constructor 
but there is no official and robust way to determine which one the parameter is at runtime.


micwallace:
Closing, included in https://github.com/TokenScript/token-negotiator/pull/493

PR #489, opened 2022-12-07
update un end point

nicktaras:


micwallace (approved):

oleggrib (commented):
This endpoint works same way as crypto-verify.heroku, all OK

PR #507, opened 2022-12-27
Feature/walletconnect v2

snowwhitedev:


src/wallet/Web3WalletProvider.ts
```diff
@@ -255,6 +255,84 @@ export class Web3WalletProvider {
...
+		universalWalletConnect.on("session_update", ({ topic, session }: { topic: string; session: SessionTypes.Struct }) => {
+			console.log("WC V2 EVENT", "session_updated");
+			// setSession(session);
+		},
+		);
+  
+		// Subscribe to session delete
+		universalWalletConnect.on("session_delete", ({ id, topic }: { id: number; topic: string }) => {
+			console.log("WC V2 EVENT", "session_deleted");
+			// resetApp();
```
micwallace, line 294: I presume this is when the user deletes the WC session from the wallet. In this case we should probably delete the connection and show the wallet selector UI.

micwallace:
Hey @snowwhitedev there is an issue with session persistence in this implementation. I have been trying to work out a solution but no luck yet.

nicktaras:
Screen Shot 2023-01-05 at 3 28 17 pm

Screen Shot 2023-01-05 at 3 34 23 pm

@snowwhitedev - When building the UMD module version, there is an error. Text Encoder is not a constructor. 
I think you have likely added these highlighted imports by accident. Can you review and update this.

Thanks.

micwallace:
@nicktaras @snowwhitedev I've added a commit to fix the wallet persistence issue, as well as the text encoder issue pointed out by Nick. Seems to be working OK, at least with trust-wallet.

src/wallet/Web3WalletProvider.ts
micwallace: @snowwhitedev @nicktaras I've added the logic to disconnect on our end when we receive this event, however there is a bug in the WC provider at the moment that prevents this event from firing.

micwallace (dismissed):
Awesome, thanks @snowwhitedev

nicktaras (approved):
Looks good! Great work @snowwhitedev.

As you've shared this library is at an early stage with little support for many wallets yet. This is ok. We'll create a follow up ticket from here to ensure we keep support for the WC tech.

nicktaras:
@snowwhitedev - I saw that Miccy and I have both approved this but it hasn't been merged.

Please merge whenever ready.

snowwhitedev:
@nicktaras @micwallace  this branch was rebased by Miccy. Could you guys approve this again?

nicktaras (approved):

PR #511, opened 2023-01-05
Feature/expose token store

nicktaras:
Updated tokenStore: TokenStore visibility from private to protected. Allowing for extensibility / readability of this class.

Until this is updated, @ts-ignore can be used to allow access to this class instance within client. 

Changes include changes from main branch.


micwallace (approved):
Looks good to me @nicktaras

PR #514, opened 2023-01-09
Fix settings btn align

nicktaras:


nicktaras:
Updated src/theme/common.scss

micwallace (approved):

PR #517, opened 2023-01-09
fix: passive token persistence & redirect enhancements

micwallace:
- Added changes so that persisted tokens are used in passive mode once tokens are loaded the first time. This prevents subsequent redirects when using redirect mode. The redirect will only happen again if negotiate is called with the refreshTokens param set to true or the TTL for the cached token has expired. 
- Changed enableOffChainRedirectMode to offChainRedirectMode. By default this is undefined and redirect mode is used if the browser does not support iframe storage. It can be set to "always" or "never" for testing purposes.
- Apply the offChainRedirectMode setting to authentication as well as token loading. 

@nicktaras I believe you wanted to keep tab mode for Safari desktop & iOS? If so we'll need to modify some logic in: 
src/utils/support/getBrowserData.ts:121

The current logic is to use redirect for all iOS browsers, Safari & Brave.


nicktaras:
> * Added changes so that persisted tokens are used in passive mode once tokens are loaded the first time. This prevents subsequent redirects when using redirect mode. The redirect will only happen again if negotiate is called with the refreshTokens param set to true or the TTL for the cached token has expired.
> * Changed enableOffChainRedirectMode to offChainRedirectMode. By default this is undefined and redirect mode is used if the browser does not support iframe storage. It can be set to "always" or "never" for testing purposes.
> * Apply the offChainRedirectMode setting to authentication as well as token loading.
> 
> @nicktaras I believe you wanted to keep tab mode for Safari desktop & iOS? If so we'll need to modify some logic in: src/utils/support/getBrowserData.ts:121
> 
> The current logic is to use redirect for all iOS browsers, Safari & Brave.

This sounds great Miccy.

I'll test this out on my machine / mobile. Yep, I think keeping the tab flow would be much better to use for some of the experiences. Doing what we can to make this the best experience (accepting some initial compromises if needed on UX to provide a general solution).

Here are some thoughts on the IOS side of the solution: 

- For IOS Mobile - the tab transition for IOS is smooth for the native browsers where re-direct isn't needed
- For IOS desktop - the tab transition / re-direct isn't needed
- For IOS Wallet browsers - re-direct is needed

If we can target capabilities rather than user agent I think we can find a good solution to have control over providing the best experience across devices.

nicktaras (approved):

nicktaras (approved):

PR #518, opened 2023-01-10
fix: use UMD modules for Torus & WalletConnect to fix polyfill issues

micwallace:
This should remove the need for the vast majority of polyfills by explicitly using the UMD version of of WalletConnect & Torus. 

It's been tested on most examples after removing polyfills and is working successfully.


nicktaras (approved):
Smoke tested on examples and UMD and works correctly for WC.

Approved.

PR #519, opened 2023-01-10
Feature/event sender

nicktaras:


nicktaras:
@micwallace added changes to the eventSender method. 

This change consolidates the eventSender callback triggers into a single trigger. Including changes to enable type checking for each payload received to be emitted.

micwallace (approved):

PR #532, opened 2023-01-24
Update/tn btn sprite background

snowwhitedev:


src/theme/common.scss
```diff
@@ -793,3 +808,23 @@ li.issuer-connect-banner-tn .tokens-btn-tn {
     width: 274px !important;
   }
 }
+
+@keyframes openAnimation {
```
nicktaras, line 812: @snowwhitedev great work on adjusting this to use CSS instead of the KeyShape lib.

From running this locally, I could notice on some browsers some level of uneven frame rate in the animation.

Could we look at using css transitions to bring up the frame rate?

https://medium.com/outsystems-experts/how-to-achieve-60-fps-animations-with-css3-db7b98610108

snowwhitedev: @nicktaras  Thank you. Could you point in which browsers you can see uneven frame rate?(could you upload short video in jira?)
Regrading using transition, in my opinion transition is a bit different animation for fab button.
Using transition we can whole sprite image by inch, not one by one.
Meanwhile, I will check how to bring up frame rate when using animation attribute/
Thanks.

nicktaras: Thanks Henderson. I noticed the animation on Chrome and Safari. I can add screen recordings soon of the original and new. 

Using transition in CSS will give us access to the GPU and maximise the frame rate we can achieve via CSS.

To do this, we would need to wrap up the sprite element inside a div with an overflow of hidden e.g. 

‘’’’
<div style=“overflow: hidden; width: 1 frame px”>
   
</div> 
‘’’’

moving the sprite left and right with transition css. 

Alternatively using JS can be used to give the same performance as using key shape.

snowwhitedev:
@nicktaras  Hi Nick. I checked css transition, it shows unexpected behavior. I am sending short video in discord.
Btw the main point to use GPU is will-change property.
So I added this property and committed again
Could you review again?
Thanks.

nicktaras (approved):

micwallace (approved):

PR #542, opened 2023-02-09
updated iframe height, fix for safari

nicktaras:


nicktaras:
The changes address these two issues:

1ef304d7-ae03-4ef1-bc84-dc3ae6d60ba9

373db2ff-668b-458f-8878-7821f0196a86

nicktaras:
This is a lower priority peer review task compared to:

Attestation fixes and ERC20.

micwallace:
@nicktaras Any reason the formatting in package.json has changed?

Also I am really not a fan of setting width: 101%. It will end up causing more issues like a scrollbar appearing when it shouldn't. 

Did you check box-sizing property to see if it's set correctly? Usually this will be the cause of the gap. That or possibly a border where there shouldn't be.

micwallace:
@nicktaras commit rebased with staging, ready to merge.

nicktaras:
PR on hold until the attestation flow works again.

micwallace (approved):
Confirmed that this shouldn't cause scrollbar issues.

PR #546, opened 2023-02-10
re: disconnect event catch from MM && WalletConnevtV1

snowwhitedev:


nicktaras (approved):

nicktaras:
This looks good @snowwhitedev, thanks for cleaning this logic up.

PR #551, opened 2023-02-14
re: chain switch regression

snowwhitedev:
This PR is for TKS-723

### DIFF
--- a/src/client/index.ts
+++ b/src/client/index.ts
@@ -377,7 +377,12 @@ export class Client {
 		if (this.ui) {
 			autoOpenPopup = this.tokenStore.hasUnloadedTokens()
 
-			if (this.ui.viewIsNotStart() && this.tokenStore.hasUnloadedIssuers()) this.enrichTokenLookupDataOnChainTokens()
+			if (this.ui.viewIsNotStart() && this.tokenStore.hasUnloadedIssuers()) {
+				this.enrichTokenLookupDataOnChainTokens()
+			} else {
+				// Main screen may be shown so it requires an update event though tokens are already loaded for the updated issuers
+				this.triggerUiUpdateCallback(UIUpdateEventType.ISSUERS_LOADED)
+			}
 		} else {
 			this.createUiInstance()
 			await this.ui.initialize()
--- a/src/wallet/Web3WalletProvider.ts
+++ b/src/wallet/Web3WalletProvider.ts
@@ -181,12 +181,14 @@ export class Web3WalletProvider {
 
 		// @ts-ignore
 		provider.provider.on('chainChanged', (_chainId: any) => {
-			console.log('chainChanged ==>', _chainId)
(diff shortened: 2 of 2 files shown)


micwallace:
There's still issues here @snowwhitedev. The original issue is fixes somewhat but it has caused flow-on issues.

1. Only the first token loads when switching chain
2. Refresh button fails after chain is switched

I've worked this out to be two main issues:
1. The disconnect event is fired by MetaMask when chain switch is delayed
2. The main view does not receive the event to re-render the view after issuers are updated. 

To fix this:

Remove cache clear and reload for chainChanged event, it's not needed (we keep tokens cached for a while except account change or wallet disconnect)
```
		// @ts-ignore
		provider.provider.on('chainChanged', (_chainId: any) => {
			this.registerNewWalletAddress(accounts[0], _chainId, providerName, provider)

			this.saveConnections()

			this.emitNetworkChange(_chainId)
		})

```

Add a condition to skip disconnect event if it is related to Metamask chain switch. Luckily Metamask tells us this:
```
		// @ts-ignore
		// walletconnect
		provider.provider.on('disconnect', (reason: any) => {
			if (reason?.message && reason.message.indexOf('MetaMask: Disconnected from chain') > -1)
				return;

			this.client.disconnectWallet()
		})
```

In index.ts activeNegotiationStrategy we need to emit ISSUERS_LOADED event to the UI, even if we don't have new issuers to load. This is because previously loaded issuers could be re-negotiated:
```
			if (this.ui.viewIsNotStart() && this.tokenStore.hasUnloadedIssuers()) {
				this.enrichTokenLookupDataOnChainTokens()
			} else {
				// Main screen may be shown so it requires an update event though tokens are already loaded for the updated issuers
				this.triggerUiUpdateCallback(UIUpdateEventType.ISSUERS_LOADED)
			}
```

Once these patches are in place it should work as expected. It should not return to the wallet screen and instead it should reload tokens/issuers if required. If issuers are already loaded it will use the cached data and simply re-render the view.

micwallace:
Also I should note that we use duplicated (one per chain) collections for the ecommerce example. We listen for the network-change event in the example code and call negotiate() with new issuers. This is not really a common use case but we needed it for this example.

micwallace (changes requested):
Please see last comment for requested changes

snowwhitedev:
@micwallace  I agree with you for now. Btw I asked @nicktaras about that we are fetching all issuers data regardless of connected network. In that sense I reloaded tokens when switching chain.
Anyhow I agree with you at the moment. Could you review again and approve PR?

micwallace (approved):

micwallace:
Done. We only need to know the wallet address to load all EVM tokens. The idea is the user specifies the tokens that they are interested in (across different EVM chains) and we return them all. Otherwise the user would have to switch to each chain. The chainID is in the issuer config.

For this examples we wanted to support the same tokens via different chains to allow the user to mint either Goerli or Mumbai, but it's not a normal use case.

PR #553, opened 2023-02-16
Fix: various 2.3 issues

micwallace:
Fixed:
- TKS-628 Redirect mode causes exception before page reload
- TKS-736 Fix tokens automatic load not triggering due to stale tokenStore
- TKS-737 Firefox permission dialog not clickable (Windows & Android). Firefox should use redirect mode
- Fix addTokenViaMagicLink returning void in redirect mode & improve outlet error handling (relates to TKS-628)


nicktaras (approved):
Looks great!

Reviewed across Mac, Safari and Chrome (UMD and ES6 module versions)

PR #557, opened 2023-02-21
revert breaking change wallet connection interface

nicktaras:


src/client/index.ts
```diff
@@ -856,8 +856,8 @@ export class Client {
 	async eventSender(eventName: 'tokens', data: EventSenderTokens)
 	async eventSender(eventName: 'token-proof', data: EventSenderTokenProof)
 	async eventSender(eventName: 'tokens-selected', data: EventSenderTokensSelected)
-	async eventSender(eventName: 'connected-wallet', data: EventSenderConnectedWallet)
-	async eventSender(eventName: 'disconnected-wallet', data: EventSenderDisconnectedWallet)
+	async eventSender(eventName: 'connected-wallet', EventSenderConnectedWallet)
```
micwallace, line 859: This is incorrect, as type is specified as a parameter name. 

To fix:
1. Revert these lines
2. Change EventSenderConnectedWallet to a named union type rather than an interface: 
`export type EventSenderConnectedWallet = WalletConnection | null`

micwallace (changes requested):
@nicktaras Please remove unnecessary indent changes in index.html & package.json.

Additionally, can you please make sure that your IDE options/plugins aren't auto-formatting any files. 
Prettier/eslint is only setup to format files in src directory, so it must be some automatic thing outside that.

src/client/index.ts
nicktaras: Thanks @micwallace. As confirmed via Discord, please could you help address this interface/type issue raised.
Adding directly to this PR is fine / most efficient for us addressing this.

micwallace (approved):

PR #558, opened 2023-02-21
Feature: view transitions

micwallace:
Implement the following view transitions that can be applied by passing the options.viewTransition property to the UI.updateUI function or specifying it in the viewOverrides config options:

- slide-in-left
- slide-in-right
- slide-in-top
- slide-in-bottom

Inbuilt views have been updated to use the left/right transition.


micwallace:
Do not merge. 2.4 only.

nicktaras (approved):

PR #560, opened 2023-02-22
Fix: url query namespace

micwallace:
…(TKS-770)

- Add namespace prefix to URL query hash parameters to ensure our params don't conflict with 3rd party website that may use the hash query.
- Ensure request parameters are kept completely separate from callback URL The callback for the negotiator client site is now enveloped within the outlet callback. This will help avoid conflicts and prevent duplicate parameters.
- Rework function to remove callback URL parameters once processed.

Note: This requires a small bug fix to attestation.id to ensure that request parameters are not included with the callback URL. Please see https://github.com/AlphaWallet/attestation.id/pull/196


micwallace:
Do not merged until v2.4 is started

src/utils/index.ts
```diff
@@ -142,3 +143,27 @@ export const tokenRequest = async (query: string, silenceRequestError: boolean)
 		throw new Error(`HTTP error.`)
 	}
 }
+
```
nicktaras, line 146: Could you write a unit test for this logic, testing pass and fail scenarios.

Thanks.

src/outlet/index.ts
```diff
@@ -184,7 +186,12 @@ export class Outlet {
 						this.dispatchAuthCallbackEvent(issuer, null, e.message)
 					}
 
-					document.location.hash = ''
+					document.location.hash = removeUrlSearchParams(this.urlParams, [
+						'attestation',
+						'requestSecret',
+						'address',
+						'wallet',
+					]).toString()
```
nicktaras, line 194: Should this include the URLNS as the last parameter?

nicktaras:
> Do not merged until v2.4 is started

Let's review when this can be deployed and any impacts e.g. if this is brought into a project during development or during launch.

Thoughts are it might be best for us to integrate sooner than later to catch any issues.

src/outlet/index.ts
micwallace: No that parameter has a default value.

src/utils/index.ts
micwallace: @nicktaras Unit tests added

micwallace:
@nicktaras Let me chase up Hu and see if I can get the attestation.id changes merged first.

nicktaras:
> @nicktaras Let me chase up Hu and see if I can get the attestation.id changes merged first.

Thanks great point Miccy. Can you help open a conversation for us to gather an update from Hu.

nicktaras:
Thanks for the update on this one @micwallace.

Can you take a look at the unit tests for this, and we can merge in.

nicktaras (dismissed):

nicktaras (approved):

PR #569, opened 2023-02-27
re: fix duplicated local storage for off chain tokens

snowwhitedev:
This PR includes the fix of the ticket 
https://smarttokenlabs.atlassian.net/browse/TKS-813

FYI: The reason of this ticket issue is that we save magic link(encode tokens) in local storage regardless metadata match.

Btw we need to fix this one hard-coded part :)
https://github.com/TokenScript/token-negotiator/blob/fix/duplicated-offchaintokens-localstorage/src/outlet/index.ts#L262


src/outlet/index.ts
```diff
@@ -224,11 +232,44 @@ export class Outlet {
...
+				}
+
+				const decodedTokenData = decodeToken(
+					tokenData,
+					this.tokenConfig.tokenParser,
+					this.tokenConfig.unsignedTokenDataName,
+					false,
+				)
+				// hard-coded for temporary solution
+				if (decodedNewToken['ticketIdNumber'] !== decodedTokenData['ticketIdNumber']) {
```
micwallace, line 262: As tickets may be stored for multiple conferences (events), we need to look at devconId as well as ticketIdNumber for determining uniqueness.

I suggest you build a unique ID based on both these values, as well as the legacy ticketIdString:
const id = `${decodedToken.devconId}-${decodedToken.ticketIdNumber ?? decodedToken.ticketIdString}`;

Might be best to put it in a separate function to ensure consistency. Once this is done you can use this ID to determine uniqueness rather than just the ticketIdNumber.

index.html
```diff
@@ -19,22 +19,41 @@
       var theme = 'dark';
       function init() {
             window.negotiator = new negotiator.Client({
-                type: 'active',
```
micwallace, line 22: Could you please revert all changes to index.html? You can keep your version for testing just don't commit it.

src/outlet/index.ts
```diff
@@ -224,11 +232,44 @@ export class Outlet {
 		const { tokenUrlName, tokenSecretName, tokenIdName, itemStorageKey } = this.tokenConfig
 
 		try {
-			const tokens = readMagicUrl(tokenUrlName, tokenSecretName, tokenIdName, itemStorageKey, this.urlParams)
+			const newToken = readMagicUrl(tokenUrlName, tokenSecretName, tokenIdName, itemStorageKey, this.urlParams)
```
micwallace, line 235: Could you please add a simple unit test for this function? Or move some of the logic out into a separate function to make it easier to add a unit test without mocking localStorage? 

Maybe something like: shouldTokenBeReplaced(oldToken, newToken)?

micwallace (changes requested):
Hey @snowwhitedev I've requested a few changes here. 

Apart from that great work on separating out some of the functions. The outlet code definitely needs a bit of a refactor.

src/outlet/index.ts
snowwhitedev: Yeah, Good point. This is what I missed in the first PR. I thought of this after this PR. I will fix it in next PR

index.html
snowwhitedev: yes

src/outlet/index.ts
snowwhitedev: Okay, I separated a new function (readTokenFromMagicUrl) and added unit test.
But readMagicUrl is not used anywhere now and readTokenFromMagicUrl is used only in readMagicLink function

micwallace (approved):

micwallace:
@nicktaras Your review & approval is required here. Thanks mate.

nicktaras:
@micwallace can you please help give sign off on this when ready. Thanks.

micwallace (approved):

PR #570, opened 2023-02-28
off chain duplicate filter

nicktaras:


nicktaras:
Hi @micwallace, @snowwhitedev 

Please see this solution. I feel this is a lighter way to implement this change applying the filter at the point the tokens are to be sent to the client site. Although let's review and implement the right solution. 

Note: I've added unit tests to the duplication logic added, however because this is the first time it's been linted with the Prettier rules Sonar explains this is failing.

Thanks guys.

micwallace:
Closing as we have gone with Hendersons initial solution of preventing duplicates being added in the first place.

PR #576, opened 2023-03-02
feat: 🎸 display only relevant wallet providers

oleggrib:
✅ Closes: TKS-768


src/client/index.ts
```diff
@@ -250,6 +250,30 @@ export class Client {
 		)
 	}
 
+	public evmAvailable() {
```
micwallace, line 253: @oleggrib let's make this a single function that takes the blockchain as a parameter. Avoid repetition like so:

```
public hasIssuerForBlockchain(blockchain: 'evm' | 'solana' | 'flow') {
		return (
			this.config.issuers.filter((issuer: OnChainTokenConfig) => {
				// EVM should always be active when we have off-chain attestations as it's used for UN challenge signing
				if (blockchain === 'evm' && !issuer.onChain) return true

				// Defaults to evm if blockchain isn't specified and is an onchain token
				return (issuer.blockchain ? issuer.blockchain.toLowerCase() : 'evm') === blockchain
			}).length > 0
		)
	}
```

src/client/views/select-wallet.ts
```diff
@@ -38,36 +39,58 @@ const flowSVG = `
...
+			? `<button class="wallet-button-tn" data-wallet="SafeConnect" aria-label="Safe connect wallet button">
+			  ${safeConnectSVG}
+			  <p>Safe Connect</p>
+			</button>`
+			: ''
+
+		const browser = getBrowserData();
 		const MetaMaskButton =
-			typeof window.ethereum !== 'undefined'
+			(browser.anyMetamask && evmOrOffChain)
```
micwallace, line 54: It's probably more reliable to use window.ethereum rather than explicitly checking for metamask as other non-metamask providers may exists. i.e. does this break Brave browser wallet support?

nicktaras:
Thanks for this implementation @oleggrib, and comments added @micwallace.

This has been planned for version 2.5.0 - let's hold on this change unless we find it's needed for the following 2.4.0 release.

micwallace:
@nicktaras I think this is an important fix. We shouldn't be showing flow wallet with no flow tokens configured. If @oleggrib can respond to my requests I'm happy to get it merged in 2.4

src/client/index.ts
oleggrib: great update! added it as commit

src/client/views/select-wallet.ts
oleggrib: as for me  - TN will show inaccurate icons for brave, opera and lot of mobile wallets if it will display Metamask, while no Metamask installed.
Lets allow WalletConnect to interact with other wallets.

micwallace: This happens in alot of dapps though. For instance in alphawallet. Will this prevent injected provider from being used in alphawallet?

oleggrib: no, it just hides MetaMask  icon for non-Metamask browsers. 
It doesnt disable use of windows.ethereum
user can connect ANY wallet (metamask too) through WalletConnect

micwallace: @oleggrib yes but in that case the user has to pair wallet connect. I think it's better to have both options, even if the icon is wrong. We can always fix the icon and title of the wallet in a later version. 

In essence, let's fix it later rather than removing existing functionality.

oleggrib: I have added icons for multiple wallets, so if window.ethereum defined then appropriate icon will be displayed.
in brave browser: 
- if metamask installed and enabled then metamask icon appear
- if brave wallet enabled then Brave icon appear

also added icons for:
- MEW
- trustWallet
- statusWallet
- alphaWallet
- imToken

micwallace: @oleggrib you are a legend!

micwallace (approved):

nicktaras (approved):

PR #577, opened 2023-03-02
fix: 🐛 error handle for Flow popup close

oleggrib:
✅ Closes: TKS-739

### DIFF
--- a/src/wallet/Web3WalletProvider.ts
+++ b/src/wallet/Web3WalletProvider.ts
@@ -366,9 +366,14 @@ export class Web3WalletProvider {
 			const fcl = flowProvider.getFlowProvider()
 
 			fcl.currentUser.subscribe((currentUser) => this.flowSubscribe(fcl, currentUser))
-			fcl.authenticate()
+
+			await fcl.currentUser.authenticate()
+			let user = await fcl.currentUser.snapshot()
+
+			if (!user.addr) throw new Error('No user address after authenticate()')
 		} catch (e) {
 			console.error('error ==>', e)
+			this.client.getUi().showError('Flow wallet connection error.')
 		}
 		return ''
 	}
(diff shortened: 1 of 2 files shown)


src/wallet/Web3WalletProvider.ts
```diff
@@ -366,9 +366,14 @@ export class Web3WalletProvider {
...
 			fcl.currentUser.subscribe((currentUser) => this.flowSubscribe(fcl, currentUser))
-			fcl.authenticate()
+
+			await fcl.currentUser.authenticate()
+			let user = await fcl.currentUser.snapshot()
+
+			if (!user.addr) throw new Error('No user address after authenticate()')
 		} catch (e) {
 			console.error('error ==>', e)
+			this.client.getUi().showError('Flow wallet connection error.')
```
micwallace, line 376: UI may not always be initialized in passive mode.
Can we remove the try/catch as we want the exception to be propagated to the client. We should avoid accessing any UI specific functions in this class.

src/wallet/Web3WalletProvider.ts
```diff
@@ -366,9 +366,14 @@ export class Web3WalletProvider {
 			const fcl = flowProvider.getFlowProvider()
 
 			fcl.currentUser.subscribe((currentUser) => this.flowSubscribe(fcl, currentUser))
```
micwallace, line 368: Are we able to make this function async rather than using a flowSubscribe callback? All wallet provider functions should register the provider and return wallet address.

micwallace (changes requested):

src/wallet/Web3WalletProvider.ts
oleggrib: done

src/wallet/Web3WalletProvider.ts
oleggrib: done

micwallace (approved):
Thanks @oleggrib for the quick turnaround on this

micwallace (approved):

PR #579, opened 2023-03-02
feat: add disconnect function for attestation issuers

snowwhitedev:


snowwhitedev:
This PR is the solution  for Nick's AC in TKS-777
I added some comments in the jira ticket

src/client/index.ts
```diff
@@ -265,13 +269,20 @@ export class Client {
...
+			nextUI = 'wallet'
+			this.attestationDisconnected = false
+		}
+
 		this.tokenStore.clearCachedTokens()
 		this.eventSender('connected-wallet', null)
 		this.eventSender('disconnected-wallet', null)
 		if (this.ui) {
-			this.ui.updateUI('wallet', { viewName: 'wallet' })
+			this.ui.updateUI(nextUI as ViewType, { viewName: nextUI })
```
nicktaras, line 285: Thanks for this change.

nicktaras: From testing this as you'd mentioned there is an endless loop, where the token page cannot be re-loaded which we'll need to address for this change. 

Screen Shot 2023-03-03 at 11 53 52 am

nicktaras: Could we look at applying a rule that does something like this:

- If Auto Load is on. 
- The first time they navigate to the Main view we auto load tokens.
- If they have disconnected from the wallet to start over, we do not try autoloading the tokens this time around.

This I feel is the best UX and also should remove this issue.

cc: @micwallace

nicktaras: Hi guys, please ignore the bug above. This is a general but to be fixed outside of this task. 

Where when testing with autoLoadTokens false, this is the same behaviour.

cc: @snowwhitedev, @micwallace

nicktaras: I've created a review PR towards this suggestion basing the changes on top of what you've done to this point in the PR.

https://github.com/TokenScript/token-negotiator/commit/997c60b7acf9cfa4ce101709c6e775776b048460

nicktaras: https://github.com/TokenScript/token-negotiator/compare/for-review-only/disconnect-wallet-flow-fixes

micwallace: @nicktaras agreed. Ideally any change to the UI should be done by the UI view class. i.e. the disconnect button click handler should determine if the view needs to go back to the wallet. The only exception to this is in the disconnect-wallet event in Web3WalletProvider.

micwallace:
Hey @snowwhitedev there is a few issues that aren't covered in this PR. 

Nick & I reviewed and came up with this solution that let's the UI view class handle any updates that need to be done to the UI when disconnectWallet is called, rather than that function interacting with the UI directly:

https://github.com/TokenScript/token-negotiator/pull/580

This will be more robust and is required for BrandConnector UI, where disconnecting the wallet won't automatically navigate to the wallet connect screen.

nicktaras:
Closing this PR.

snowwhitedev:
@nicktaras  @micwallace  could you guys approve this PR again?

micwallace:
@snowwhitedev no I'm not approving it because you basically just reverted all my changes 😂 I had rebased the commits, and forced pushed it.

micwallace (approved):

micwallace:
@snowwhitedev all sorted mate. Just let me know if you added anything extra.

PR #582, opened 2023-03-06
chore: comment out magic link wallet support

micwallace:


nicktaras:
@micwallace thanks for creating this PR, please can you for now remove the code. Keeping it safely inside a Feature Branch.

Where we can re-introduce this when the requirement is confirmed. Finalising the changes with a poly-fill update to complete this item.

nicktaras (approved):

PR #589, opened 2023-03-07
Fix: TKS 853 & 854

micwallace:
- Fix injected wallet from showing when not available (needed typeof for strict comparison)
- Fix view container width causing toggles to be cut off


snowwhitedev (approved):
Looks good. 
Btw it seems there's an issue in my local network configuration, so I did not smoke test for mobile icon issue. I think we can test it in public deployed version.
Token toggle works good,

PR #592, opened 2023-03-08
fix: 🐛 fixed WC2 flow

oleggrib:
Fixes:
- WC2 stuck on Reject
- WC2 receive empty address
- WC2 double login

### DIFF
--- a/src/client/index.ts
+++ b/src/client/index.ts
@@ -276,12 +276,16 @@ export class Client {
 
 	async negotiatorConnectToWallet(walletType: string) {
 		let walletProvider = await this.getWalletProvider()
+		try {
+			let walletAddress = await walletProvider.connectWith(walletType)
 
-		let walletAddress = await walletProvider.connectWith(walletType)
+			logger(2, 'wallet address found: ' + walletAddress)
 
-		logger(2, 'wallet address found: ' + walletAddress)
-
-		return walletAddress
+			return walletAddress
+		} catch (e) {
+			logger(2, 'walletProvider.connectWith error: ', e)
+		}
+		return ''
 	}
 
 	async enrichTokenLookupDataOnChainTokens() {
--- a/src/client/index.ts
+++ b/src/client/index.ts
@@ -276,16 +276,7 @@ export class Client {
 
 	async negotiatorConnectToWallet(walletType: string) {
 		let walletProvider = await this.getWalletProvider()
-		try {
-			let walletAddress = await walletProvider.connectWith(walletType)
-
-			logger(2, 'wallet address found: ' + walletAddress)
-
-			return walletAddress
-		} catch (e) {
-			logger(2, 'walletProvider.connectWith error: ', e)
-		}
-		return ''
+		return await walletProvider.connectWith(walletType)
 	}
 
 	async enrichTokenLookupDataOnChainTokens() {
--- a/src/client/views/select-wallet.ts
+++ b/src/client/views/select-wallet.ts
@@ -274,8 +274,12 @@ export class SelectWallet extends AbstractView {
 		)
 
 		try {
-			await this.client.negotiatorConnectToWallet(wallet)
+			let walletAddress = await this.client.negotiatorConnectToWallet(wallet)
 			this.ui.dismissLoader()
+			
+			// if user rejected connect to account
+			if (!walletAddress) throw new Error("Wallet didn't connect")
+
 			if (this.params?.data?.connectCallback) {
 				this.params?.data?.connectCallback()
 			} else {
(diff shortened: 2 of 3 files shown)


src/client/index.ts
```diff
@@ -276,12 +276,16 @@ export class Client {
...
+		try {
+			let walletAddress = await walletProvider.connectWith(walletType)
 
-		let walletAddress = await walletProvider.connectWith(walletType)
-
-		logger(2, 'wallet address found: ' + walletAddress)
+			logger(2, 'wallet address found: ' + walletAddress)
 
-		return walletAddress
+			return walletAddress
```
micwallace, line 284: Can we remove all the try catches? It's better to let the exception propagate so when there are errors the message is specific and has the correct stack trace.

src/wallet/Web3WalletProvider.ts
```diff
@@ -97,15 +97,20 @@ export class Web3WalletProvider {
...
 
 		if (this[walletType as keyof Web3WalletProvider]) {
-			// @ts-ignore
-			// TODO: this address is null for the flow network
-			// Actual connected address is get in flowSubscribe function
-			const address = await this[walletType as keyof Web3WalletProvider](checkConnectionOnly)
-			logger(2, 'address', address)
+			try {
+				// @ts-ignore
+				// TODO: this address is null for the flow network
```
micwallace, line 102: I think this TODO is outdated after recent changes to the Flow integration. Can you check and remove if so?

src/wallet/Web3WalletProvider.ts
```diff
@@ -97,15 +97,20 @@ export class Web3WalletProvider {
...
 
-			this.saveConnections()
-			this.emitSavedConnection(address)
-			return address
+				this.saveConnections()
+				this.emitSavedConnection(address)
+				return address
+			} catch (e) {
+				logger(2, 'Provider connect Error: ', e)
+				throw new Error(e.message)
```
micwallace, line 112: Let's remove this try/catch, or a the very least use "throw e;" rather than "throw new Error(e.message)" to retain the original stacktrace.

src/client/views/select-wallet.ts
```diff
@@ -274,8 +274,12 @@ export class SelectWallet extends AbstractView {
 		)
 
 		try {
-			await this.client.negotiatorConnectToWallet(wallet)
+			let walletAddress = await this.client.negotiatorConnectToWallet(wallet)
 			this.ui.dismissLoader()
+			
+			// if user rejected connect to account
+			if (!walletAddress) throw new Error("Wallet didn't connect")
```
micwallace, line 281: If the exception propagates, this shouldn't be necessary.

micwallace (dismissed):

src/client/index.ts
oleggrib: done

src/wallet/Web3WalletProvider.ts
oleggrib: yes, outdated, removed.

src/wallet/Web3WalletProvider.ts
oleggrib: done

src/client/views/select-wallet.ts
micwallace: @oleggrib Good point but this can be thrown in Web3WalletProvider class so it works for all UIs (i.e. BC). 
Maybe at the end of Web3WalletProvider.connectWith() method?

micwallace (approved):

micwallace:
Hey @oleggrib I've approved but noticed the build is failing. It looks like "connected" is not a property on UniversalProvider. 
I also tested locally and these changes break the ability for WCv2 to have session persistence. On each page refresh the user will need to re-pair their wallet. 

I've looked around the codebase of UniversalProvider and there doesn't seem to be any easy way of determining connection state. 
This is why I used the original solution of calling the "enable" function to determine if there's an existing connection. 

I found this possible solution but it may not be the best. Another possible solution is what was done for WCv1
```
if (checkConnectionOnly) {

		universalWalletConnect
			.enable()
			.then(() => {
				const provider = new ethers.providers.Web3Provider(universalWalletConnect, 'any')

				resolve(this.registerProvider(provider, 'WalletConnectV2'))
			})
			.catch((e) => {
				reject('Not connected')
			})

	} else {

		universalWalletConnect
			.connect({
					...
```

micwallace:
@oleggrib Maybe we should test out like this: https://docs.walletconnect.com/2.0/javascript/sign/dapp-usage

Note comment: "// Optionally: pass a known prior pairing (e.g. from `signClient.core.pairing.getPairings()`) to skip the `uri` step."

if checkConnectionOnly is true and uri is not null then we can know the session is dead.

oleggrib:
Hi, @micwallace , thanks for finding this issue. I habe added correct detection of active session and corrected disconnect flow. WC1 and WC2 shoul connect/reconnect/disconnect good now. Could you check it?

micwallace:
Working awesome, thanks @oleggrib

micwallace (approved):

PR #600, opened 2023-03-14
Possible fix for vite issue importing WCv2 UMD module

micwallace:
@futantan can you please test this out when you get a chance?


futantan:
hi @micwallace I would like to confirm the purpose of the test:
1. we will use UMD version
2. we need to test if this branch can contribute to smaller bundle size

Is that correct?
another options is build a ES module from this branch, in order to test this, we need:
1. publish a NPM package
2. or using `npm link` in my local

which one is correct?

futantan (approved):

futantan:
LGTM

PR #605, opened 2023-03-17
feat: add reload issuers example in index.html

snowwhitedev:


snowwhitedev:
This PR does not include any essential changes. I added one button and function which makes TKN to reload updated issuers

micwallace (approved):

PR #612, opened 2023-03-21
Fix: prettier eslint format conflict

micwallace:
Hey @nicktaras this should fix most of the formatting conflicts between eslint & prettier.
And increases printWidth to be in line with CSS setting. This makes the HTML template literals a clearer to read but can be changed back if you want.


nicktaras:
Thanks @micwallace I'll take a look and aim to merge this in.

snowwhitedev (approved):

PR #616, opened 2023-03-22
feat: added new event hook tokens-loaded

snowwhitedev:


micwallace:
@snowwhitedev Although it wasn't mentioned in the AC, I'm not sure that @nicktaras intended the return data to be the total number of tokens. Nick can you please clarify the AC to specify what this events data should contain?

It also needs work to define what "successfully loaded" is. Does this mean if a token fails to load then this event should not fire? Or is it what I presumed - fire once after a loading process is complete.

nicktaras (dismissed):
Hi Henderson, could you update the README with the hook.

Also, how about we nest the return value inside an object e.g.

{ count: length }

This in case we wish to provide more details / change the interface later, it would be more flexible for us.

nicktaras:
> 

Thanks @snowwhitedev for thinking about this and your inputs too @micwallace to see what's best and confirm the missed AC here. 

I agree with your idea Henderson, and have added a comment / idea around this for your review. 

I'll make sure the AC is up to scratch on this. Apologies for times when this is lacking some of this level of detail.

src/client/index.ts
```diff
@@ -646,6 +648,7 @@ export class Client {
 		// 	return arg;
 		// }
 		this.eventSender('tokens', tokens)
+		this.eventSender('tokens-loaded', Object.keys(tokens).length)
```
micwallace, line 651: @snowwhitedev This will always return the number of issuers configured in TN rather than the number of collections that have tokens, see TokenStore.getCurrentTokens method. Notice it returns all issuers with an empty array for unloaded issuers, or issuers that have 0 tokens. 

Have a look at TokenStore.getTotalTokenCount in PR 604, maybe you can create a new method TokenStore.getIssuersWithTokensCount which does a similar thing.

micwallace (changes requested):

src/client/index.ts
snowwhitedev: @micwallace  I guess we have misunderstanding about the meaning of returning data. My expectation is to return the number of collections which are loaded regardless whether user has any tokens in that collection or not.
Let's say easily, there's only Perion collection in issuers config, Whenever I have any perion tokens or not, once Perion collection is loaded, `tokens-loaded` event returns 1. This is my expectation.
cc: @nicktaras

nicktaras:
Hi @snowwhitedev & @micwallace,

From reviewing your questions raised, I've revised the AC of this ticket to give detailed requirements (with more AC).

Great idea to return data at this point Henderson, I've made an adjustment to the AC around this to what I feel will cater to various use cases for the hook itself.

If there are any changes needed to the AC from taking a look, let us review and adjust the ticket together / or smaller changes please adjust the AC accordingly.

snowwhitedev:
> Hi @snowwhitedev & @micwallace,
> 
> From reviewing your questions raised, I've revised the AC of this ticket to give detailed requirements (with more AC).
> 
> Great idea to return data at this point Henderson, I've made an adjustment to the AC around this to what I feel will cater to various use cases for the hook itself.
> 
> If there are any changes needed to the AC from taking a look, let us review and adjust the ticket together / or smaller changes please adjust the AC accordingly.

Yes, I will commit today with updated read me and test cases

micwallace (approved):
Awesome thanks @snowwhitedev

micwallace (approved):

PR #617, opened 2023-03-24
Chore: minimize bundle size

micwallace:
Optimizations to reduce UMD bundle entrypoint file. The extra dynamic imports should also optimize usages of the NPM package depending on bundler & configuration used (i.e. bundler setup to support chunking & tree shaking). 

- Always use LocalOutlet in client rather than normal Outlet.
- Create AbstractOutlet to share methods required by both LocalOutlet & Outlet.
- Dynamically load several classes when required:
  - LocalOutlet
  - Web3WalletProvider (this was already being dynamically import but also statically imported for type usage. I've updated this to import type only)
  - Authentication methods ticketZKProof & signedUNChallenge
  - Authenticator library
- Cleanup and remove some redundant functionality in Outlet.
- Separate webpack entrypoint files for Client & Outlet.

### DIFF
--- a/src/client/auth/ticketZKProof.ts
+++ b/src/client/auth/ticketZKProof.ts
@@ -1,10 +1,8 @@
 import { AbstractAuthentication, AuthenticationMethod, AuthenticationResult } from './abstractAuthentication'
 import { AuthenticateInterface, OffChainTokenConfig, OnChainTokenConfig } from '../interface'
 import { OutletAction, Messaging } from '../messaging'
-import { Authenticator } from '@tokenscript/attestation'
 import { SignedUNChallenge } from './signedUNChallenge'
 import { UNInterface } from './util/UN'
-import { LocalOutlet } from '../../outlet/localOutlet'
 import { OutletInterface } from '../../outlet'
 import { logger } from '../../utils'
 import { shouldUseRedirectMode } from '../../utils/support/getBrowserData'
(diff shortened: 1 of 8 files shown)


nicktaras:
@micwallace please can you help to continue this PR;

- Raise the coverage of the changes made to 40% or above for the changes made, to pass the libraries quality gate. 
- Fix the conflicts with the outlet/index.ts

Thanks. 

Also from reviewing the AC you have done both the investigation and implementation in one ticket. From the initial review part of this task, could you help share the optimisation gains you've seen from the changes made in the ticket - which we can share as part of the comms when we release. 

Screen Shot 2023-03-31 at 10 42 35 am

micwallace:
@nicktaras I think Sonarcube coverage calculation is a bit strange. Seems to only use changed lines or files. I spent some time adding more tests for outlet, and it did increase Jest coverage report a fair bit but barely raised the sonarcloud figure. 
It would be pretty hard to get it higher as most of the code is related to postMessage stuff.

nicktaras:
With the 2.5.0 release delayed. We'll plan deployment of this inside 2.6.0 release to allow for time to test the changes inside the PR and to capture the gains found.

micwallace:
Before optimization: 
image

After optimization:
image

micwallace:
There are bugs due to the version compatibility changes we introduced recently. It will need to be reworked to correct these issues.

nicktaras:
> @nicktaras I think Sonarcube coverage calculation is a bit strange. Seems to only use changed lines or files. I spent some time adding more tests for outlet, and it did increase Jest coverage report a fair bit but barely raised the sonarcloud figure. It would be pretty hard to get it higher as most of the code is related to postMessage stuff.

Hi Miccy, 

Just coming back to this PR. Your point raised around unit tests has been looked into and found that the unit tests we're inaccurate. Where it was found the unit tests themselves we're being included by Sonar but not by Jest into the coverage scores. This has now been fixed, where I'll next help re-base your PR (let me know if you'd prefer to do this etc) and we can review and merge next week.

nicktaras:
Will re-base and merge for version 2.8.0 of TKN to ensure there is enough window to record the gains made and ensure nothing regressed from this code optimisation. Where 2.7.0 release is planned over a short window.

micwallace:
Probably safer to rewrite this into a fresh branch to avoid merge conflicts. Shouldn't be too difficult.

PR #625, opened 2023-03-30
Fix/ts config fix

nicktaras:
@snowwhitedev, @BankiLouis,

This change is to address an issue found when handing over on the Brand Connector tasks to Hu's team.

Where it was found that the TS process settings block the ability to extend JS classes. 

To resolve this issue, I've reverted the TS config back to the state it was since version 2.3.0, and run tests with Tantan across Brand Connector and our examples to help give confidence this change will enable us to release this change. But more thorough testing will be needed from Jean/QA team etc.

Pro's
- This change gives the benefit of supporting older browsers. 
- Unblocks BC and any other products that extend out library classes with custom behaviours

Con's
- NPM package size will increase to 140kb

Moving forward we should be able to fix this through the right config settings - but feel this is the safest most logical approach at this time. Giving us time to explore this, considering the various options we have to improve this part of our product.

Thanks team 

cc: @tcglam


snowwhitedev (approved):

snowwhitedev:
@nicktaras  I approved, but I guess you should resolve conflicts first.

snowwhitedev (approved):

PR #626, opened 2023-04-03
feat: 🎸 added sign/verify for flow and solana, fixes

oleggrib:
sign/verify UN for Flow and Solana
fixed default param name id->mail, according to the attestation.jar
fixe iframe height bug , child iframe limited to 100vh

### DIFF
--- a/src/client/auth/signedUNChallenge.ts
+++ b/src/client/auth/signedUNChallenge.ts
@@ -2,6 +2,7 @@ import { AbstractAuthentication, AuthenticationMethod, AuthenticationResult } fr
 import { AuthenticateInterface, OffChainTokenConfig, OnChainTokenConfig } from '../interface'
 import { SafeConnectProvider } from '../../wallet/SafeConnectProvider'
 import { UN, UNInterface } from './util/UN'
+import { logger } from '../../utils'
 
 export class SignedUNChallenge extends AbstractAuthentication implements AuthenticationMethod {
 	TYPE = 'signedUN'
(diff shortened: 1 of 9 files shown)


src/client/auth/signedUNChallenge.ts
```diff
@@ -15,25 +16,30 @@ export class SignedUNChallenge extends AbstractAuthentication implements Authent
 		let web3WalletProvider = await this.client.getWalletProvider()
 
 		// TODO: Update once Flow & Solana signing support is added
-		if (web3WalletProvider.getConnectedWalletData('evm').length === 0) {
+		let connection = web3WalletProvider.getSingleSignatureCompatibleConnection();
```
micwallace, line 19: Hey @oleggrib we need to use the connection for the specific token which is being authenticated, rather than the first token. This will allow multi-wallet & multi-blockchain support going forward. 

To do this you can use:
```
web3WalletProvider.getWalletProvider(tokens[0].walletAddress)
```
It will throw an exception if the wallet is no longer connected.

micwallace (changes requested):

src/client/auth/signedUNChallenge.ts
oleggrib: in current flow we make UN signature and after that start PROOF flow, so at the moment of start getTokenProof() we dont have info about required wallet.
ticket allows any wallet, required wallet will be known only when emailAttestation received.
in that case we can decode attestation and get user address @micwallace

micwallace: We use signedUNChallenge for on-chain token verification as well as off-chain. 
So you are right in off-chain case, tokens[0].walletAddress won't be set.
Maybe we check for this value and get the associated wallet if it's provided, otherwise we get first EVM wallet. 

There's no reason to have Solana/Flow for off-chain at the moment, until we have smart contracts on those chains.

oleggrib: UN signature can be used for authentication with wallet. 
UN separate from on-chain, off-chain flow
so UN can be signed with evm/solana/flow wallet and verified with API and with TN itself.

I think this PR is OK to add ability to sign/verify with solana/flow. We can add limitations in other PR @micwallace 
What do you think?

micwallace: Okay I will fix when we introduce multi-wallet UI.

micwallace (approved):

PR #632, opened 2023-04-05
updated attestation lib

nicktaras:


nicktaras:
Updated the attestation lib to use the latest version.

oleggrib (approved):

PR #634, opened 2023-04-06
Update readme for solana issuers

snowwhitedev:
Add one warning for the case of solana


nicktaras (approved):
Can we adjust this copy to not include, 'warning'. I feel this can just be a comment

PR #655, opened 2023-04-27
added cursor style to buttons for outlet

nicktaras:
Really minor change added to align the button styles to all use, cursor: pointer for the TKN UX.


snowwhitedev (approved):
Looks good

PR #669, opened 2023-05-10
Fix/keep open popup connecting

snowwhitedev:
This PR is for the ticket https://smarttokenlabs.atlassian.net/browse/TKS-1043

### DIFF
--- a/src/client/ui.ts
+++ b/src/client/ui.ts
@@ -191,7 +191,6 @@ export class Ui implements UiInterface {
 				})
 
 				document.addEventListener('click', () => {
-					console.log('document event listener')
 					this.closeOverlay()
 				})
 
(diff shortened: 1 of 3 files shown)


nicktaras:
Hi @snowwhitedev 

From smoke testing this, I found that the user has to click twice to close the overlay when declining a connection. 

Could we try to apply this flow to this code change:

- User opens overlay and clicks let's go
- Connect to wallet
- Declines connection and wants to next close the TKN
- Clicks once outside of the overlay region / or overlay button to close the TKN

snowwhitedev:
> Hi @snowwhitedev
> 
> From smoke testing this, I found that the user has to click twice to close the overlay when declining a connection.
> 
> Could we try to apply this flow to this code change:
> 
> * User opens overlay and clicks let's go
> * Connect to wallet
> * Declines connection and wants to next close the TKN
> * Clicks once outside of the overlay region / or overlay button to close the TKN

Yes, I will look into

snowwhitedev:
@nicktaras  I committed again. Could you look into again? 🙏

nicktaras (approved):

nicktaras (approved):

PR #675, opened 2023-05-16
added return to resolving promise from a user action

nicktaras:
Review of change to return promise from user action, when accepting or declining to connect with a token issuer website to read attestations.

### DIFF
--- a/src/outlet/index.ts
+++ b/src/outlet/index.ts
@@ -360,9 +360,8 @@ export class Outlet {
 					min_height: this.tokenConfig.whitelistDialogHeight,
 				})
 			})
-		} else {
-			return 'user-accept'
 		}
+		return 'user-accept'
 	}
 
 	prepareTokenOutput(filter: any) {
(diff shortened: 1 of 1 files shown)


src/outlet/index.ts
```diff
@@ -360,6 +360,8 @@ export class Outlet {
 					min_height: this.tokenConfig.whitelistDialogHeight,
 				})
 			})
+		} else {
```
micwallace, line 363: @nicktaras else is redundant since the if statement returns a promise, can you please remove it?

nicktaras: Thanks for taking a look @micwallace could you run this code on your side too against the examples.

Where from testing my side, without the else the function will automatically return sending the tokens before permission is provided.

nicktaras: Apologies my side on this. 

I'd forgotten we are now producing a second copy of the attestation inside the ticket issuer (main site) which I hadn't removed from storage, giving the impression that we could load tokens without permission granted.

I've resolved removing the else and re-tested and works perfectly.

snowwhitedev (approved):

PR #700, opened 2023-06-02
Feat: multi-attestation outlet

micwallace:
- Update ticketStorage to support multiple attestations
- Add import of multiple attestations (updated attestation lib to support this)
- Implement client/outlet changes to handle requests for multiple attestations
- Update whitelist to display all requested collections and allow the user to deny specific collections


nicktaras (approved):
Changes reviewed and tested locally. 

One issue detected for Brave added via Jira to address as a seperate fix. 

Approved.

PR #701, opened 2023-06-02
Fix/multi outlet issuer active

nicktaras:


src/client/views/select-issuers.ts
```diff
@@ -103,11 +103,11 @@ export class SelectIssuers extends AbstractView {
 			return
 		}
 
-		this.populateIssuers()
-
-		let tokensListElem = this.tokensContainer.getElementsByClassName('token-list-container-tn')[0]
-
-		this.tokenListView = new TokenList(this.client, this.ui, tokensListElem, { ...this.params })
+		setTimeout(() => {
```
micwallace, line 106: Hey @nicktaras 

I found a better solution that fixes the core of the issue:

src/client/index.ts:454
Change:
```
if (!refresh && tokens != null) continue
```
to:
```
if (tokens != null){
	onComplete(issuerKey, tokens);
	continue;
}
```

Then above the loop, around src/client/index.ts:450 add the following:
```
if (refresh)
	this.tokenStore.clearCachedTokens();
```

nicktaras: Thanks Miccy. Looks good to me - tested and works as expected.

PR #704, opened 2023-06-05
re: fix brave wallet icon size

snowwhitedev:


nicktaras:
Screen Shot 2023-06-06 at 4 39 47 pm

The change to the alignment looks great 🙏

nicktaras (approved):
@snowwhitedev - what do you think about the naming convention

Brave (Solana)

I was thinking for consistency this should likely be Phantom (Brave)

Please update if you think this is a better name.

Approved.

PR #713, opened 2023-06-09
Chore/align main staging

nicktaras:


src/outlet/ticketStorage.ts
```diff
@@ -226,28 +226,30 @@ export class TicketStorage {
...
-				ticket.tokenId = this.getUniqueTokenId(await this.decodeTokenData(ticket.type, ticket.token))
-			}
-
-			if (ticket.tokenId === tokenRecord.tokenId) {
-				collectionTickets[index] = tokenRecord
-				this.ticketCollections[collectionHash] = collectionTickets
-				this.storeTickets()
-				return
+	private async updateOrInsertTicket(collectionHash: string, tokenRecord: StoredTicketRecord): Promise<boolean> {
+		{
```
oleggrib, line 230: hi, @nicktaras , why do we need extra brackets?

src/version.ts
```diff
@@ -1,2 +1,3 @@
 // modified by build process.
 export const VERSION = '2.8.0'
```
oleggrib, line 2: it says in Readme changes 2.7.0...2.7.1 , but VERSION says 2.8.0
is it OK, @nicktaras

README.md
```diff
@@ -497,6 +497,19 @@ Changing the theme.
...
 
+### Tokens Updated Hook
+
+Detect when new tokens have been added to the Token Negotiator during an applications lifecycle (for off chain tokens at this time).
+
+```javascript
+
+// temporary solution likely to change in the next major release version.
+document.body.addEventListener("tokensupdated", () => {
+	console.log("Tokens updated event fired!!");
```
oleggrib, line 508: we have to add new negotiate(), because tokens list changed after that action

nicktaras: Thanks @oleggrib, I think this should be ok for the README.

nicktaras: Where the end developer may wish to use this event for another task. E.g. Read the new list of tokens and navigate them to a gated page location.

src/outlet/ticketStorage.ts
nicktaras: Thanks for spotting this. It was an accidental merge conflict when aligning yours and Miccy's changes.

src/version.ts
nicktaras: This is ok for staging - we can update when we move to release 2.8.0

micwallace (approved):

PR #722, opened 2023-06-16
added multi-auth updates to lib api

nicktaras:
Adjusted the authenticate library method to support old single attestations and new multi.

authenticate is now a co-ordination method - 

is single token - it will return from authenticateToken()
is multi token - it will return many from athenticateMutilple() which uses, authenticateToken() per token

### DIFF
--- a/package-lock.json
+++ b/package-lock.json
@@ -1,12 +1,12 @@
 {
   "name": "@tokenscript/token-negotiator",
-  "version": "2.7.0",
+  "version": "2.8.0",
   "lockfileVersion": 2,
   "requires": true,
   "packages": {
     "": {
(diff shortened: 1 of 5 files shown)


nicktaras:
@micwallace could you please review these changes.

Focus has been placed on:

- Backward compatibility for the Authenticate method
- The output interface, providing details that will allow developers to easily detect if all proof's were returned as whole, allowing this to be flexible for future updates / backwards compatibility (with the aim that we can later phase out single attestation flows and always return an array to simplify this for users)

Thanks

nicktaras:
Closing this PR. It wasn't able to meet the further and more complex requirements of this task. Where wip-multi-auth branch was created and now in PR.

PR #730, opened 2023-06-21
Feat/wip multi auth

nicktaras:
Hi @micwallace, @oleggrib,

Please see this PR towards adding multi-token support to Token Negotiator. 

This enables;

- redirect, tab and iframe modes to support multi token authentication requests for a single issuer outlet.
- tested via Chrome, Safari, FF and Brave for passive and active flows

For any testing towards review please use the feat/wip-mulit-auth branch in examples:

- Hotel Bogota Active, select the tokens to auth, then click 'book now'
- Ticket Issuer Outlet, select 2 tokens to authenticate

I will spend a few more hours on this today, to reduce/remove any added duplications from the changes added. But the core implementation will stay the same. 

Thanks guys.


README.md
```diff
@@ -481,8 +483,28 @@ negotiator.authenticate({
...
+```javascript
+/**
+ * @param {{issuer: String, unsignedToken: Object}[]} tokens object array of tokens
+ */
+negotiator.authenticate([{
+	issuer,
+	unsignedToken,
+}]);
+
+negotiator.on("proof", ({ issuers, issuersValidated }) => {
```
oleggrib, line 504: Hi, @nicktaras , I think better use different event name, like "multiproof", because 2 events with same name and different return data can confuse people and can force mistakes in code

README.md
```diff
@@ -481,8 +483,28 @@ negotiator.authenticate({
...
 	// the proof will be received here (valid or failed)
+	// issuers: { issuerName: [ token, ... ] }
+	// issuersValidated: [ 'issuerName1', 'issuerName2', '...' ]
```
oleggrib, line 507: @nicktaras , issuers object will containe only valid data, right? in that case all missings issuers invalid. so "issuersValidated" property contains duplicated data, do I miss something?

src/client/auth/signedUNChallenge.ts
```diff
@@ -1,18 +1,14 @@
...
+import { AbstractAuthentication, AuthenticationMethod, AuthenticationResult, AuthenticationMethodUN } from './abstractAuthentication'
 import { AuthenticateInterface, OffChainTokenConfig, OnChainTokenConfig } from '../interface'
 import { SafeConnectProvider } from '../../wallet/SafeConnectProvider'
 import { UN, UNInterface } from './util/UN'
 import { logger } from '../../utils'
 
-export class SignedUNChallenge extends AbstractAuthentication implements AuthenticationMethod {
+export class SignedUNChallenge extends AbstractAuthentication implements AuthenticationMethodUN {
 	TYPE = 'signedUN'
 	private static DEFAULT_ENDPOINT = 'https://attestation-verify.tokenscript.org/un'
```
oleggrib, line 9: @nicktaras would be great to move DEFAULT_ENDPOINT to the default config, to stay all default values in single place.

src/client/auth/abstractAuthentication.ts
```diff
@@ -19,6 +19,20 @@ export interface AuthenticationMethod {
...
+	getTokenProof(request: AuthenticateInterface): Promise<AuthenticationResult>
+}
+
+export interface MultiAuthenticateInterface {
+	options?: any
+}
+
+export interface AuthenticationMethodMulti {
+	TYPE: string
+	getTokenProofMulti(userTokens: Array<any>, request: MultiAuthenticateInterface): Promise<AuthenticationResult>
```
oleggrib, line 33: @nicktaras interface describes type of "request", but implementation use "any", better use specific type, it can help avoid mistakes and IDE helps to code.

src/client/auth/abstractAuthentication.ts
```diff
@@ -19,6 +19,20 @@ export interface AuthenticationMethod {
 	): Promise<AuthenticationResult>
 }
 
+export interface AuthenticationMethodUN {
+	TYPE: string
+	getTokenProof(request: AuthenticateInterface): Promise<AuthenticationResult>
+}
+
+export interface MultiAuthenticateInterface {
+	options?: any
```
oleggrib, line 28: @nicktaras are you going to add additional properties to that interface? if no then better avoid "options" and describe types if its possible

src/client/auth/abstractAuthentication.ts
```diff
@@ -19,6 +19,20 @@ export interface AuthenticationMethod {
 	): Promise<AuthenticationResult>
 }
 
+export interface AuthenticationMethodUN {
+	TYPE: string
+	getTokenProof(request: AuthenticateInterface): Promise<AuthenticationResult>
```
oleggrib, line 24: @nicktaras interface AuthenticationMethod use same method name, but different input params. better change method name or make same params. This difference can lead to errors, better avoid that.

src/client/auth/ticketZKProofMulti.ts
```diff
@@ -0,0 +1,148 @@
...
+	private messaging = new Messaging()
+
+	// NOTE: A limitation at this time is that the user can only authenticate tokens
+	// that use the same issuer origin, email and it's underlying config (see: firstCollectionFound via getTokenProofMulti).
+
+	async getTokenProofMulti(userTokens: MultiTokenInterface[], request: any): Promise<AuthenticationResult> {
+		const firstCollectionFound = Object.keys(userTokens)[0]
+		const tokenCollectionConfig = userTokens[firstCollectionFound].requestTokens[0]
+		const issuerCollectionConfig = userTokens[firstCollectionFound].issuerConfig
+		let redirectMode: false | string = this.getRedirectModeSetting(tokenCollectionConfig, request)
```
oleggrib, line 27: @nicktaras better disable option to force redirect. it depends on browser only. for testing we can use Brave or Safari and Chrome. simplier code = less errors and easier to update

src/client/auth/ticketZKProofMulti.ts
```diff
@@ -0,0 +1,148 @@
...
+import { DEFAULT_EAS_SCHEMA, TokenType } from '../../outlet/ticketStorage'
+import { EAS_RPC_CONFIG } from '../../core/eas'
+
+export class TicketZKProofMulti extends AbstractAuthentication implements AuthenticationMethodMulti {
+	TYPE = 'ticketZKProofMulti'
+
+	private messaging = new Messaging()
+
+	// NOTE: A limitation at this time is that the user can only authenticate tokens
+	// that use the same issuer origin, email and it's underlying config (see: firstCollectionFound via getTokenProofMulti).
```
oleggrib, line 21: @nicktaras would be great to foreach tokens and if it use different origins or configs then throw new Error(). it will help to dont forget that limitation

src/client/auth/ticketZKProofMulti.ts
```diff
@@ -0,0 +1,148 @@
...
+			}
+			let unRes = await unChallenge.getTokenProof(request)
+			useEthKey = unRes.data as UNInterface
+		}
+		const useEthKeyAddress = useEthKey ? useEthKey.address : ''
+		const address = request.address ? request.address : useEthKeyAddress
+		const wallet = request.wallet ? request.wallet : ''
+		let output
+		if (new URL(issuerCollectionConfig.tokenOrigin).origin === window.location.origin) {
+			output = await this.authenticateLocally(userTokens, issuerCollectionConfig, address, wallet, redirectMode)
```
oleggrib, line 43: @nicktaras could you make same params order for authenticateLocally(userTokens, issuerCollectionConfig...) and authenticateCrossOrigin(issuerCollectionConfig, userTokens...) to avoid problems in future?

src/client/interface.ts
```diff
@@ -78,6 +78,12 @@ export interface WalletOptionsInterface {
 	}
 }
 
+// issuer: { requestTokens, issuerConfig }
+export interface MultiTokenInterface {
+	requestTokens: any[]
+	issuerConfig: OffChainTokenConfig
```
oleggrib, line 84: @nicktaras , could you define type of requestTokens, because hard to understand why do we need "issuerConfig" as separate param.
in getTokenProofMulti() I can see code:
```const tokenCollectionConfig = userTokens[firstCollectionFound].requestTokens[0]```
could you help me to understand difference between issuerConfig and tokenConfig?

File: src/outlet/index.ts [hunk 11]
```diff
@@ -121,25 +120,60 @@ export class Outlet {
 						const attestationBlob = this.getDataFromQuery('attestation')
 						const attestationSecret = '0x' + this.getDataFromQuery('requestSecret')
```
oleggrib, line 121: @nicktaras , could you add check if query params exist , non-empty and contains data in correct format?

src/outlet/index.ts
```diff
@@ -121,25 +120,60 @@ export class Outlet {
...
-							requesterURL.hash = params.toString()
-
-							window.location.href = requesterURL.href
-
-							return
+						let issuerConfig = this.getIssuerConfigById(issuer) ?? this.tokenConfig
+
+						let useToken
+
+						let localStorageAuthRequest = JSON.parse(localStorage.getItem('token-auth-request'))
```
oleggrib, line 127: @nicktaras could you add 'token-auth-request' to the default config or make as Constant, defined in some config file? because its value duplicated in 2 positions, someone can change one of them by accident

src/outlet/index.ts
```diff
@@ -214,6 +327,48 @@ export class Outlet {
...
+						output.issuersValidated.push(key)
+					}
+					for (const token of tokensObj[key].requestTokens) {
+						const issuer = this.getIssuerConfigById(key)
+						const ticketRecord = await this.ticketStorage.getStoredTicketFromDecodedToken(
+							createIssuerHashArray(issuer),
+							token.unsignedToken,
+						)
+						let authHandler = new AuthHandler(issuer, ticketRecord, tokensObj[key].unsignedToken, address, wallet, redirect, this, evtid)
+						let tokenProof = await authHandler.authenticate()
```
oleggrib, line 353: @nicktaras in Chrome it case if email attestation is not ready mutiple simultaneous authenticate() will open multiple iframes or rewrite each other. anyway problem can happen.

README.md
nicktaras: I'll change this to "issuersProcessed" to make it clearer to the end developer these are the issuers that were processed as part of the authentication.

nicktaras: I'll leave this conversation open to review back with you on this.

index.html
```diff
@@ -26,22 +26,65 @@
...
-                55: "MFYwEAYHKoZIzj0CAQYFK4EEAAoDQgAEAGJAHCiHbrCNAY9fAMdom4dGD6v/KkTIgRCkwLCjXFTkXWGrCEXHaZ8kWwdqlu0oYCrNQ2vdlqOl0s26/LzO8A==",
+                55: "MIIBMzCB7AYHKoZIzj0CATCB4AIBATAsBgcqhkjOPQEBAiEA/////////////////////////////////////v///C8wRAQgAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAEIAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAHBEEEeb5mfvncu6xVoGKVzocLBwKb/NstzijZWfKBWxb4F5hIOtp3JqPEZV2k+/wOEQio/Re0SKaFVBmcR9CP+xDUuAIhAP////////////////////66rtzmr0igO7/SXozQNkFBAgEBA0IABGMxHraqggr2keTXszIcchTjYjH5WXpDaBOYgXva82mKcGnKgGRORXSmcjWN2suUCMkLQj3UNlZCFWF10wIrrlw=",
               },
               base64attestorPubKey:
                 "MIIBMzCB7AYHKoZIzj0CATCB4AIBATAsBgcqhkjOPQEBAiEA/////////////////////////////////////v///C8wRAQgAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAEIAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAHBEEEeb5mfvncu6xVoGKVzocLBwKb/NstzijZWfKBWxb4F5hIOtp3JqPEZV2k+/wOEQio/Re0SKaFVBmcR9CP+xDUuAIhAP////////////////////66rtzmr0igO7/SXozQNkFBAgEBA0IABL+y43T1OJFScEep69/yTqpqnV/jzONz9Sp4TEHyAJ7IPN9+GHweCX1hT4OFxt152sBN3jJc1s0Ymzd8pNGZNoQ=",
               ticketIssuesUrlWebsitePrivateKey:
```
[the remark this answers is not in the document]
nicktaras: Thanks, I'll update this and make sure all still works ok.

src/client/auth/abstractAuthentication.ts
nicktaras: Thanks, I must have missed this. I'll use the interface.

src/client/auth/abstractAuthentication.ts
nicktaras: Yes, the intention is to move global settings for multi-tokens into this param / interface in the next steps of the epic.

README.md
nicktaras: I would like to in version 3.0.0 provide the same interface for Single and Multi. Where 2.8.0 and 2.9.0 will use proof and for version 3.0.0 there will only be multi - regardless of having 1 or more tokens we can use the same interface and delete much of our code.

nicktaras: Having flows for single and multi are too complicated I feel - there should be as fewer processes and learnings for the end developer and users.

src/client/auth/signedUNChallenge.ts
nicktaras: I agree, let's to streamline this PR / task move this to a Jira ticket within this EPIC.

I'll create a Jira ticket capturing this task to address.

nicktaras: https://smarttokenlabs.atlassian.net/browse/TIKAT-292

src/client/auth/ticketZKProofMulti.ts
nicktaras: I'll split this into a new Jira ticket. For us to remove this, but also for the examples that uses this as a means of testing + any of the BDD/E2E testing suite that uses this trigger to simulate the tests.

nicktaras: https://smarttokenlabs.atlassian.net/browse/TIKAT-293 - Moved to a seperate task for us to take on. Thanks for the suggestion to look at doing this.

src/outlet/index.ts
nicktaras: From testing this in the console, it appears to do everything nicely in order. This is a very complex process for the browser to make here - My preference after this MVP solution is to move this to attestation.id to resolve all email attestations and allow the TKN to process them - this will be a much more user friendly experience also (keeping the user in one place at a time).

src/outlet/index.ts
nicktaras: Updated to use **LocalStorageMessaging.TOKEN_AUTH_REQUEST**  in these locations

src/outlet/index.ts
nicktaras: Let's move this to a seperate Jira task, where this is a change related to previous work. Thanks for picking this up to look into.

nicktaras: https://smarttokenlabs.atlassian.net/browse/TIKAT-294

src/client/interface.ts
nicktaras: Thanks. I found this is using the wrong interface. Ill update this.

src/client/auth/abstractAuthentication.ts
nicktaras: Can we confirm the changes needed. 

I'd added the Un interface to simplify the UN export class SignedUNChallenge extends AbstractAuthentication implements AuthenticationMethodUN

Which had a different interface to AuthenticationMethod whilst developing the changes for multi-auth. 

I'll take a further look at this today, in case I've not understood the first time reading this correctly.

src/client/auth/ticketZKProofMulti.ts
nicktaras: Thanks for this point. This is what has been done so far to address this, so at this point the tokens should only be from the same issuer origin (per batch request).

Via client getMultiRequestBatch() - we assign the tokens with the same tokenOrigin into batches.
authRequestBatch.offChain[issuerConfig.tokenOrigin]

As you mentioned we don't check the config settings, but think this should be enough for now to ensure they are using the correct issuer, letting the following process take hold of if the ticket is valid or not?

Where if I understand your thought on checking this config at this point, may help us to reject the multi-request earlier?

README.md
nicktaras: Updated.

src/outlet/index.ts
nicktaras: Thoughts are we can change this to trigger a singleton class, and have the method receive the given params,
but this would be more likely to have code re-writing each other in a scenario like this - my understanding is that Miccy had taken this approach upfront to allow for many classes to sit in memory, such as Auth handler and some others that use this idea - instantiating class state and having methods with no params to use the class state data etc?

src/client/auth/abstractAuthentication.ts
oleggrib: I mean AuthenticationMethod and AuthenticationMethodUN have mathod **getTokenProof**, but those methods have different input params, it can lead to errors

src/client/auth/abstractAuthentication.ts
oleggrib: in case if you are going to add additional data as properties of **options** then this interface is redundant. 
it ok if some inteface have struture like {options, data, config} . is it our case?
and please update type of **options** from any to some specific. much easier to read/review code when exact types used.

src/client/auth/abstractAuthentication.ts
oleggrib: could you create interface for userTokens ? 
thanks

src/client/auth/ticketZKProofMulti.ts
oleggrib: I mean that we can forgot this limitation and get error
in case if we limit code according to some rule then we have to set simple check and in rule broken them throw an error with verbose description.
in cese if someone use incompable config then error will happen and we will know the reason. @nicktaras

src/client/auth/ticketZKProofMulti.ts
oleggrib: we can easy change user-agent in testing , no need to pass specific param. @nicktaras 
thank for a task!

README.md
oleggrib: hi, @nicktaras , can't we use Object.keys(issuers) instead of issuersProcessed/issuersValidated ?

README.md
oleggrib: maybe you forgot to push your updates, because in manual it still use 
negotiator.on("proof",... in both scenarios, but input data are different.

src/client/interface.ts
oleggrib: in MultiTokenInterface we allow multiple tokens, but single issuerConfig, is it OK? what if someone send tokens for different issuers?

src/outlet/index.ts
oleggrib: its easy to fix by running authenticate() one by one, syncroniously. just add for() and let tokenProof = await authHandler.authenticate() inside loop @nicktaras

README.md
nicktaras: Sorry it looks like I'd not responded to the correctly. This was intentional, so for version 3.0.0 we can move to one implementation for proof - to return an array.

README.md
nicktaras: Yep, I'll remove this, it's not essential. I had added this initially from past experience where dev teams liked to do it like this rather than use object keys or for key in etc.

src/client/auth/ticketZKProofMulti.ts
nicktaras: Miccy is working on the solution to this limitation at the moment, shared the PR for this. I'll close this thread @oleggrib

src/client/auth/abstractAuthentication.ts
micwallace: I removed this interface, we will need to add a type for options but this will be the same for all tokens in most cases.

src/client/interface.ts
micwallace: @oleggrib I've updated this to send issuer hashes to the outlet. So outlet can determine it's own config from the hash provided. This should solve this issue.

src/outlet/index.ts
micwallace: Done. Auth handler has been renamed to attestationIdClient and will store ID attestations in local storage to support redirect mode.

PR #750, opened 2023-07-12
Feat: improvements to multi-token authentication

micwallace:
Hey @nicktaras here is that PR I mentioned. It's quite extensive but it gets rid of a lot of code duplication and cleans up some tech-debt.

Here's a summary:
- Move AuthHandler to AttestationIdClient and repurpose to handle retrieval and storage of identifier attestations. This prevents the need for subsequent redirects to attestation.id after initial creation as ID attestations are now stored in Outlet until they are expired.
- Standardize multi-token auth request and response into interfaces that are used for both local outlet and remote outlet authentication. This reduces the size of the URL parameter ensuring only collection hashes are sent in the URL, rather than the whole config object. 
- Remove duplication in attestation.id callback switch statement. Instead we can capture ID attestation and then reuse logic from Outlet.sendTokenProof and Outlet.sendMultiTokenProof.
- Remove redundant dispatchAuthCallbackEvent in Outlet. Local outlet attestation callback is now handled in the client, with no requirement for any Outlet instance.
- Outlet now extends LocalOutlet in order to make use of the authenticateMany function across Client and Outlet.

There's still a few improvements that could be make client side to reduce duplication. For instance, sharing more of the same logic in TicketZKProofMulti class and Client.processAttestationIdCallback function. 

Note: At the moment signed UN + ZKProof is not supported in redirect mode. We have never found a good use-case for it so I think it's okay to leave it for now, but it could be added back fairly easily.


nicktaras (approved):

nicktaras:
Looks good. Thanks Miccy.

PR #752, opened 2023-07-13
center aligned icon

nicktaras:


nicktaras:
Adjusted the view box slightly to align the AW icon.

PR #758, opened 2023-07-19
Switch default UN creation api to common-api

JustinFeng:
As part of the `common-api` work, we migrated the UN creation API from `winding-tree`, and we made a change regarding the HTTP method from Get to Post, it's mainly because this endpoint is more like creating a new UN than reading an existing UN, but open to discuss.

__Note:__ We were planning to terminate the winding-tree backend later when this migration is released, especially considering the UN endpoint is only internally used by us, but please feel free to raise your concern if there is other unexpected impact (as the previous TN package will still have the default UN endpoint pointing to winding tree), cheers cc @foxgem


nicktaras (approved):

nicktaras:
@JustinFeng approved.

PR #759, opened 2023-07-20
Feat: arbitrary eas schema

micwallace:
This PR allows TN to store and read attestations that have an arbitrary schema. 

It defines a new Outlet config interface that allows the issuer to define the schema UID and field config in order to decode custom EAS tokens. 

ZKProof attestations (useToken) can also be created & verified as long as the schema defines a field called "commitment" which is the Pedersen commitment of the users email address. Apart from that, the user can define the fields used to generate a composite token ID that can be used to overwrite existing tokens when reissuing attestations. 

Example config can be found in this examples PR: https://github.com/TokenScript/token-negotiator-examples/pull/229


src/outlet/ticketStorage.ts
```diff
@@ -53,11 +55,22 @@ export type DecodedToken = DecodedTokenData & {
 }
 
 export interface DecodedTokenData {
-	devconId: string
+	tokenId?: string
+	eventId?: string
 	ticketIdNumber?: number
```
oleggrib, line 60: Hi, @micwallace . As we do big updates then lets decide to use one field  ticketIdNumber(numeric, must be unique) or ticketIdString(string, hash it to make numeric value for NFT_ID) and name it ticketId
what do you think ? 
@nicktaras

micwallace: For EAS it is always a string type. This only has implications for smart contracts usage which is done at the application level, so I think this is a good idea. For old ASN attestations with ticketIdNumber we can just convert it into a string.

src/outlet/interfaces.ts
```diff
@@ -1,4 +1,5 @@
 import { TokenType } from './ticketStorage'
+import { AbiFieldTypes } from '@tokenscript/attestation/dist/eas/EasTicketAttestation'
```
nicktaras, line 2: What are your thoughts for us to use a naming convention here. In case we need to introduce various AbiFieldTypes? e.g. for tokens beyond ticketAttestation. However this is ok for now, we can iterate changes here etc if preferred.

src/outlet/ticketStorage.ts
nicktaras: Should we change ticketIdNumber to this below for correctness that this is provided in different types? Name and types.

export interface DecodedTokenData {
	tokenId?: string
	eventId?: string
	ticketId?: number | string"
}

src/outlet/ticketStorage.ts
```diff
@@ -68,17 +81,17 @@ export interface FilterInterface {
 	[key: string]: any
 }
 
-export const DEFAULT_EAS_SCHEMA = {
+export const DEFAULT_EAS_SCHEMA: TicketSchema = {
 	fields: [
-		{ name: 'devconId', type: 'string' },
+		{ name: 'eventId', type: 'string' },
```
nicktaras, line 86: 🙏 great this is changed to be more generic.

src/outlet/ticketStorage.ts
```diff
@@ -244,9 +261,32 @@ export class TicketStorage {
...
+			for (const fieldDef of fieldDefinition) {
+				easFieldData[fieldDef.name] = {
+					label: fieldDef.label ?? fieldDef.name,
+					value: easData[fieldDef.name],
+				}
+			}
+
+			tokenData = {
+				eventId: easData.eventId,
+				ticketIdString: easData.ticketIdString,
```
nicktaras, line 282: ticketIdString. This would relate to the comments in the other section of the PR around this naming convention / management of types.

src/outlet/ticketStorage.ts
micwallace: @nicktaras Yeah I suppose we can just let the user convert to string if it's needed.

src/outlet/interfaces.ts
micwallace: @nicktaras AbiFieldTypes are always the same, unless ethereum adds new ones but I doubt it.

nicktaras (approved):

nicktaras (approved):

PR #765, opened 2023-07-24
added ticketId to attestation schema

nicktaras:
ref branch to use a "non type" reference key for the ticketId.

### DIFF
--- a/src/outlet/ticketStorage.ts
+++ b/src/outlet/ticketStorage.ts
@@ -2,11 +2,10 @@ import { URLSearchParams } from 'url'
 import { EasTicketAttestation, TicketSchema } from '@tokenscript/attestation/dist/eas/EasTicketAttestation'
 import { KeyPair } from '@tokenscript/attestation/dist/libs/KeyPair'
 import { base64ToUint8array, createIssuerHashArray, createOffChainCollectionHash, IssuerHashMap, logger } from '../utils'
-import { uint8tohex } from '@tokenscript/attestation/dist/libs/utils'
 import { Ticket } from '@tokenscript/attestation/dist/Ticket'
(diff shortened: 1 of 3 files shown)


src/outlet/ticketStorage.ts
```diff
@@ -393,7 +397,11 @@ export class TicketStorage {
 			return parts.join('-')
 		}
 
-		return `${decodedToken.eventId}-${decodedToken.ticketIdNumber ?? decodedToken.ticketIdString}`
```
micwallace, line 396: Rather than doing this, check for ticketIdNumber or ticketIdString in the decodeTokenData function. 

```
// Add
decodedToken.ticketId = decodedToken.ticketIdString ?? decodedToken.ticketIdNumber;
delete decodedToken.ticketIdString;
delete decodedToken.ticketIdString;
// before this existing code
tokenData = this.propsArrayBufferToHex(decodedToken['ticket'])
```

micwallace (changes requested):

src/outlet/ticketStorage.ts
```diff
@@ -57,6 +57,7 @@ export type DecodedToken = DecodedTokenData & {
 export interface DecodedTokenData {
 	tokenId?: string
 	eventId?: string
+	ticketId?: string | number
 	ticketIdNumber?: number
```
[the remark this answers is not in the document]
nicktaras: Removed these from the schema. Using TS ignore in the other change to allow for this.

src/outlet/ticketStorage.ts
nicktaras: Applied this change. Handling any scenarios when a legacy attestation format is used for ticketId

micwallace (approved):

PR #768, opened 2023-07-24
Feature/active interface change

nicktaras:
Adjusted Active and Passive modes selected tokens interface towards version 3.0.0.

- Simplifies the interface for the library, with one hook for tokens selected
- Reduces the level of nested data for active mode (removed the selectedTokens key)
- Updated the examples to align with this change to the lib interface (PR: https://github.com/TokenScript/token-negotiator-examples/pull/230)

### DIFF
--- a/src/client/index.ts
+++ b/src/client/index.ts
@@ -466,9 +466,7 @@ export class Client {
 		}
 
 		if (this.config.autoEnableTokens && Object.keys(this.tokenStore.getSelectedTokens()).length)
-			this.eventSender('tokens-selected', {
-				selectedTokens: this.tokenStore.getSelectedTokens(),
-			})
+			this.eventSender('tokens-selected', this.tokenStore.getSelectedTokens())
 
 		if (openPopup || (this.config.uiOptions.autoPopup === true && autoOpenPopup)) this.ui.openOverlay()
 	}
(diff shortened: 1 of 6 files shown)


src/client/index.ts
```diff
@@ -466,9 +466,7 @@ export class Client {
 		}
 
 		if (this.config.autoEnableTokens && Object.keys(this.tokenStore.getSelectedTokens()).length)
-			this.eventSender('tokens-selected', {
-				selectedTokens: this.tokenStore.getSelectedTokens(),
-			})
+			this.eventSender('tokens-selected', this.tokenStore.getSelectedTokens())
```
micwallace, line 469: @ I think it's still a good idea to keep the tokens nested, since it allows us to add other fields in the future without a breaking change. How about the event data looking like this: 
```
{
  tokens: 
  // We could add these later without breaking change
  tokenCount: number
  collectionCount: number
}
```

micwallace (approved):

src/client/index.ts
micwallace: Also please update eventSender tokens-selected interface with the new format.

micwallace, line 469: Let's keep the tokens data nested so we can allow backward compatibility without breaking changes:
```
{
   tokens: { [collectionId: string]: (DecodedToken|any)[] } // Any union for now until we formalise on-chain token interface
   // then we can add these later
   tokenCount: number,
   collectionCount: number
}
```

Also please update the event sender interface to match the new structure.

nicktaras: Updated the PR to keep the namespace selectedTokens.

PR #776, opened 2023-07-31
removal of wcv1 and fix for forced modal opening

nicktaras:
Removal of WCv1 and fix to WCv2 to allow the overlay to close when the user denies to connect with Wallet.


oleggrib (approved):
I updated WC2 config and WC connects OK

PR #787, opened 2023-08-11
migration supported added to enable migration of token storage toward…

nicktaras:
…s backward compatibility.

### DIFF
--- a/src/outlet/ticketStorage.ts
+++ b/src/outlet/ticketStorage.ts
@@ -284,6 +284,8 @@ export class TicketStorage {
 
 			tokenData.tokenId = this.getUniqueTokenId(tokenData, idFields)
 		} else {
+			console.log('decode token')
+
 			let decodedToken = new readSignedTicket(base64ToUint8array(token))
 
 			// TODO: Validate ASN.1 tokens when they are imported
@@ -404,7 +406,6 @@ export class TicketStorage {
 	}
 
 	public migrateLegacyTokenStorage(issuerConfig: OffChainTokenConfig, tokenStorageKey = 'dcTokens') {
-		console.log('migrate tokens')
 		if (!issuerConfig || !tokenStorageKey) {
 			errorHandler('Issuer config and token storage key are required for migration.', 'error', null, null, true, false)
 		}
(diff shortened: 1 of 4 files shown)


src/outlet/ticketStorage.ts
```diff
@@ -402,4 +402,19 @@ export class TicketStorage {
...
 	}
+
+	public migrateLegacyTokenStorage(issuerConfig: OffChainTokenConfig, tokenStorageKey = 'dcTokens') {
+		if (!issuerConfig || !tokenStorageKey) {
+			errorHandler('Issuer config and token storage key are required for migration.', 'error', null, null, true, false)
+		}
+		const storageData = localStorage.getItem(tokenStorageKey)
+		if (storageData) {
+			let ticketCollectionsUpdated = JSON.parse(localStorage.getItem(TicketStorage.LOCAL_STORAGE_KEY)) as unknown as TicketStorageSchema
+			const createCollectionHashes = createIssuerHashArray(issuerConfig)
```
micwallace, line 413: @nicktaras You need to create the hash for the actual attestation, not the issuerConfig, which may include several hashes.

src/outlet/ticketStorage.ts
```diff
@@ -402,4 +402,19 @@ export class TicketStorage {
 	private storeTickets() {
 		localStorage.setItem(TicketStorage.LOCAL_STORAGE_KEY, JSON.stringify(this.ticketCollections))
 	}
+
+	public migrateLegacyTokenStorage(issuerConfig: OffChainTokenConfig, tokenStorageKey = 'dcTokens') {
```
micwallace, line 406: @nicktaras You could probable just pass collectionId rather than the whole issuer config, which will already be defined.
Alternatively we could make it a static function that doesn't rely on any class properties.

src/outlet/ticketStorage.ts
nicktaras: Let's review over a call or dm thread to confirm the solution details, this implementation is inspired from other parts of the library such as ticketZKProof which using the same method to create the issuerHash array.

// line 59, ticketZKProof
const createCollectionHashes = createIssuerHashArray(issuerConfig)

src/outlet/ticketStorage.ts
nicktaras: Sounds good to improve this.

nicktaras:
Merging to staging reviewed and tested.

PR #821, opened 2023-09-11
apply promise to sync applied check

nicktaras:


oleggrib (approved):
I fixed code and now it works
Thank you.

PR #824, opened 2023-09-18
fix: prevent signed token from being deleted during authentication

micwallace:
And update single token auth to send tokenId instead of the full token in the request.


micwallace:
@nicktaras I've also upgraded attestation library to support newer EAS attestation format.

